PR #5, opened 2020-09-29
Proof Of Concept for Snippets in Eclipse using Language Server

AnkushSharma2698:
This PR handles loading and parsing of JSON snippets into a SnippetRegistry which is invoked for completion items when the user requests autocompletion options.

### DIFF
--- a/jakarta.ls/src/main/java/io/microshed/jakartals/JakartaTextDocumentService.java
+++ b/jakarta.ls/src/main/java/io/microshed/jakartals/JakartaTextDocumentService.java
@@ -91,8 +91,6 @@ public class JakartaTextDocumentService implements TextDocumentService {
 		This method is automatically called by the Language Server Client
 		provided it has provided a java-completion-computer extension on the client side.
 		*/
-		// Find a way to maintain a context Per Document
-		// We then pass this context into the CompletionItem with what the current URI has in its classpath
 		return CompletableFuture.completedFuture(Either.forLeft(
 			snippetRegistry.getCompletionItemNoContext(new Range(position.getPosition(), position.getPosition()), "\n", true)
 		));
--- a/jakarta.ls/src/main/java/io/microshed/jakartals/JakartaTextDocumentService.java
+++ b/jakarta.ls/src/main/java/io/microshed/jakartals/JakartaTextDocumentService.java
@@ -36,7 +36,9 @@ import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.HoverParams;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.LocationLink;
+import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.PublishDiagnosticsParams;
+import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.SymbolInformation;
 import org.eclipse.lsp4j.TextEdit;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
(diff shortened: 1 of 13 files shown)


jakarta.ls/src/main/java/io/microshed/jakartals/commons/SnippetRegistry.java
```diff
@@ -0,0 +1,306 @@
...
+			// LOGGER.info(snippetTypes.toString());
+			String label = snippet.getPrefixes().get(0);
+			CompletionItem item = new CompletionItem();
+            item.setLabel(label);
+            item.setDetail(snippet.getDescription());
+			String insertText = getInsertText(snippet, false, lineDelimeter);
+            item.setKind(CompletionItemKind.Snippet);
+            item.setDocumentation(Either.forRight(createDocumentation(snippet, canSupportMarkdown, lineDelimeter)));
+            item.setFilterText(label);
+			item.setTextEdit(new TextEdit(replaceRange, insertText));
```
kathrynkodama, line 188: @AnkushSharma2698  Ryan and I discussed and have come up with this solution to fix the indentation for now.  Seems that formatting the snippets based on where they are inserted is an additional feature that VS Code offers, Eclipse does not have the same support.  

For now, on the server side we want to use the range to determine the indentation.
ie. if the range looks like the following, we would use the character value to add in \t characters to our newText, such that the newText would now be something like: `@GET\n\t@Produces(MediaType.TEXT_PLAIN)\n\tpublic String ${1:methodname}() {\n\t	return "hello";\n\t}`
```
TextEdit [
  range = Range [
    start = Position [
      line = 19
      character = 1
    ]
    end = Position [
      line = 19
      character = 1
    ]
  ]
  newText = "@GET\n@Produces(MediaType.TEXT_PLAIN)\npublic String ${1:methodname}() {\n	return "hello";\n}"
]
```

Reach out with any questions!

jakarta.ls/src/main/java/io/microshed/jakartals/commons/SnippetRegistry.java
```diff
@@ -182,15 +182,43 @@ private static Snippet createSnippet(
...
+			item.setTextEdit(textEdit);
 			item.setInsertTextFormat(InsertTextFormat.Snippet);
             return item;
 		}).collect(Collectors.toList());
 	}
 
+	/**
+	 * Applies character based formatting
+	 * @param edit
+	 * @return
```
kathrynkodama, line 201: remove the `@return` from the javadoc comment since it is a void method, or have the method return the edited `TextEdit` item

jakarta.ls/src/main/java/io/microshed/jakartals/commons/SnippetRegistry.java
```diff
@@ -0,0 +1,334 @@
...
+            item.setKind(CompletionItemKind.Snippet);
+            item.setDocumentation(Either.forRight(createDocumentation(snippet, canSupportMarkdown, lineDelimiter)));
+            item.setFilterText(label);
+			item.setTextEdit(new TextEdit(replaceRange, insertText));
+            item.setInsertTextFormat(InsertTextFormat.Snippet);
+            return item;
+        }).collect(Collectors.toList());
+	}
+
+	public List<CompletionItem> getCompletionItemNoContext(final Range replaceRange, final String lineDelimeter, boolean canSupportMarkdown) {
```
kathrynkodama, line 175: lets add a javadoc comment to this method

jakarta.ls/src/main/java/io/microshed/jakartals/JakartaTextDocumentService.java
```diff
@@ -78,6 +84,20 @@ public void didClose(DidCloseTextDocumentParams params) {
...
 
+	@Override
+	public CompletableFuture<Either<List<CompletionItem>, CompletionList>> completion(CompletionParams position) {
+		/*
+		Code completion functionality for eclipse Jakarta EE.
+		This method is automatically called by the Language Server Client
+		provided it has provided a java-completion-computer extension on the client side.
+		*/
+		// Find a way to maintain a context Per Document
+		// We then pass this context into the CompletionItem with what the current URI has in its classpath
```
kathrynkodama, line 95: is this a todo?  Is the idea that the context will offer us insight into the classpath?

AnkushSharma2698: Yeah sorry that is a todo, I will remove it from the PR, just to keep it clean.

AnkushSharma2698: And yeah maintaining information about each document such as imports and such will allow us to perform direct compares with the snippet context to determine if it should be shown or not. That is how I think it'll work so far, but I will have a more solid answer once I actually start looking into context based snippet displaying

kathrynkodama: Sounds good, thanks Ankush :)

kathrynkodama (approved):
Looks good to me :) Squash your commits into 1 commit before merging please

Issue #7, opened 2020-09-29
Jakarta Servlet @WebServlet diagnostic


kathrynkodama:
Add a diagnostic to verify that classes annotated with `@WebServlet` must extend the HttpServlet class. 

https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/annotation/WebServlet.html
https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/http/HttpServlet.html

See https://eclipse-ee4j.github.io/jakartaee-tutorial/servlets004.html
*note the tutorial is referring to Jakarta EE 8, with Jakarta EE 9 the package names will change to jakarta... which is why they are slightly different in the apidocs.

kathrynkodama:
See this example of a servlet with Jakarta EE 9: https://github.com/rzgry/demo-servlet/blob/main/src/main/java/io/openliberty/sample/system/SimpleHello.java

kathrynkodama:
Closed with #23

PR #23, opened 2020-10-19
webServlet diagnostics implementation (issue #7)

PENGYUXIONG:
there are following problems currently

- when changes happened on client side, the aync call makes the diagnostic msg does not update in time.
- get signature returned format in form "Q***;" which is the unresolved signature format in string. Not sure whether it is acceptable, might need trim.

### DIFF
--- a/jakarta-eclipse/src/org/jakarta/lsp4e/JakartaLanguageClient.java
+++ b/jakarta-eclipse/src/org/jakarta/lsp4e/JakartaLanguageClient.java
@@ -1,6 +1,5 @@
 package org.jakarta.lsp4e;
 
-import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


jakarta-eclipse/src/org/jakarta/lsp4e/JakartaLanguageClient.java
```diff
@@ -73,13 +75,38 @@ public boolean isCanceled() {
...
+							boolean isWebServletAnnotated = false;
+							boolean isHttpServletExtended = false;
+							for (IAnnotation annotation : allAnnotations) {
+								if (annotation.getElementName() == "WebServlet") {
+//									System.out.println("--Annotation name: " + annotation.getElementName());
+									isWebServletAnnotated = true;
+								}
+							}
+
+							String typeExtension = type.getSuperclassTypeSignature();
```
kathrynkodama, line 97: If you use `type.getSuperclassName()`, you will receive the name instead of the typeSignature (no leading `Q` in the name)

jakarta-eclipse/src/org/jakarta/lsp4e/JakartaLanguageClient.java
```diff
@@ -73,13 +75,38 @@ public boolean isCanceled() {
...
+								isHttpServletExtended = true;
+							}
+
+							if (isWebServletAnnotated && !isHttpServletExtended) {
+								ISourceRange nameRange = JDTUtils.getNameRange(type);
+								Range range = JDTUtils.toRange(unit, nameRange.getOffset(), nameRange.getLength());
+								diagnostics.add(new Diagnostic(range, "Need to extend HttpServlet class after annotate WebServlet"));
+							}
+
 							IMethod[] methods = type.getMethods();
```
kathrynkodama, line 110: Lets remove the sample diagnostic now

PENGYUXIONG: Do you want me to remove the test diagnostic message?

I think the initial test diagnostic was removed by Raza already.

rzgry:
Merged this so that others can use code they depend on from this PR. Opened https://github.com/MicroShed/jakarta-ls/issues/26 to follow up on the delayed diagnostics.

PR #42, opened 2020-11-04
feat: created a generic filter snippet

### LINKED ISSUE #12: Jakarta Servlet Filter snippet
Add a snippet to Java files for implementing the Filter interface.

https://jakarta.ee/specifications/servlet/5.0/apidocs/jakarta/servlet/Filter.html

See:
https://eclipse-ee4j.github.io/jakartaee-tutorial/servlets006.html
https://github.com/eclipse-ee4j/jakartaee-tutorial-examples/blob/9814044413785e7b3428fd0f0259a3f7c778ef69/web/servlet/mood/src/main/java/jakarta/tutorial/mood/TimeOfDayFilter.java

*note the tutorial is referring to Jakarta EE 8, with Jakarta EE 9 the package names will change to jakarta... which is why they are slightly different in the apidocs.

AnkushSharma2698:
Signed-off-by: AnkushSharma2698 <ankush1@ualberta.ca>

I decided against adding any extra functionality in the doFilter snippet because the use cases for this Filter are extremely varied and one snippet can not possibly cover the breadth of use cases. So keeping it as a clean snippet seemed most appropriate. 

This snippet has been tested with a demo-servlet application which has Jakarta included.

Closes #12

### DIFF
--- a/jakarta.ls/src/main/resources/servlet.json
+++ b/jakarta.ls/src/main/resources/servlet.json
@@ -76,5 +76,33 @@
         "context":{
             "type": "jakarta.servlet.http.HttpServlet"
         }
-    }
+    },
+    "Jakarta Servlet WebFilter": {
+        "prefix": "servletfilter",
+        "body": [
+          "package ${1:packagename};",
+          "",
+          "/* Filter Dependencies */",
+          "import jakarta.servlet.Filter;",
+          "import jakarta.servlet.FilterChain;",
+          "import jakarta.servlet.ServletException;",
+          "import jakarta.servlet.ServletRequest;",
+          "import jakarta.servlet.ServletResponse;",
+          "import jakarta.servlet.annotation.WebFilter;",
+          "",
+          "import java.io.IOException;",
+          "",
+          "@WebFilter(filterName = \"${2:filterName}\", urlPatterns = {\"${3:urlPattern}\"})",
+          "public class ${4:className} implements Filter {",
+          "\t@Override",
+          "\tpublic void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException{",
+          "\t\tchain.doFilter(req, res);",
+          "\t}",
+          "}"
+        ],
+        "description": "Jakarta Servlet WebFilter",
+        "context":{
+            "type": "jakarta.servlet.Filter"
+        }
+      }
 }
--- a/jakarta.ls/src/main/resources/servlet.json
+++ b/jakarta.ls/src/main/resources/servlet.json
@@ -78,11 +78,10 @@
         }
     },
     "Jakarta Servlet WebFilter": {
-        "prefix": "servletfilter",
+        "prefix": "servletwebfilter",
         "body": [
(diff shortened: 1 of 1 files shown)


jakarta.ls/src/main/resources/servlet.json
```diff
@@ -76,5 +76,33 @@
...
             "type": "jakarta.servlet.http.HttpServlet"
         }
-    }
+    },
+    "Jakarta Servlet WebFilter": {
+        "prefix": "servletfilter",
+        "body": [
+          "package ${1:packagename};",
+          "",
+          "/* Filter Dependencies */",
```
kathrynkodama, line 85: I don't think we need this comment

jakarta.ls/src/main/resources/servlet.json
```diff
@@ -76,5 +76,33 @@
         "context":{
             "type": "jakarta.servlet.http.HttpServlet"
         }
-    }
+    },
+    "Jakarta Servlet WebFilter": {
+        "prefix": "servletfilter",
```
kathrynkodama, line 81: Perhaps this prefix should be `servletwebfilter`

jakarta.ls/src/main/resources/servlet.json
```diff
@@ -76,5 +76,33 @@
...
+          "import jakarta.servlet.FilterChain;",
+          "import jakarta.servlet.ServletException;",
+          "import jakarta.servlet.ServletRequest;",
+          "import jakarta.servlet.ServletResponse;",
+          "import jakarta.servlet.annotation.WebFilter;",
+          "",
+          "import java.io.IOException;",
+          "",
+          "@WebFilter(filterName = \"${2:filterName}\", urlPatterns = {\"${3:urlPattern}\"})",
+          "public class ${4:className} implements Filter {",
```
rzgry, line 96: For the class name you could try using `${TM_FILENAME_BASE}`. Ex. https://github.com/eclipse/lsp4mp/blob/master/microprofile.ls/org.eclipse.lsp4mp.ls/src/main/resources/org/eclipse/lsp4mp/snippets/mp-health.json#L15

I am not sure if it works in eclipse so u would have to give it a try.

kathrynkodama: If this works we should update the other servlet snippets as well

AnkushSharma2698: I just did some testing, and the proposed method doesnt actually work in eclipse rn. I tried messing around with templates in eclipse and that does fill in snippets with variables as we want but for whatever reason that is not extending to our current completion framework. So for now I would say we should keep snippets like this. In order to accomodate the change in this request I think i would have to restructure completionProposalComputer a bit and I am not sure about what is needed for that yet

kathrynkodama: Sounds good, thanks for opening this issue #43 Ankush, we can leave the snippets as-is for now and address this at a later date.

kathrynkodama (approved):

PR #100, opened 2021-02-22
 Change used namespace from javax.* to jakarta.* in JAX-RS snippets

berasogut:
Change the used namespace from javax.* to jakarta.* in JAX-RS resource class and method snippets.


kathrynkodama (approved):
Thanks @berasogut!

Issue #114, opened 2021-03-05
Quick fix for Jakarta Bean Validation constraints on static fields diagnostic


lamminade:
See #55

> Static fields and static methods are excluded from validation. Add a diagnostic that delivers a warning if a user tries to use a constraint annotation on a static field or method.

Quick fix should suggest removing the constraint annotation on a static field or method

lamminade:
@kathrynkodama could you verify this is the desired behaviour for the quickfix and assign to me :)

kathrynkodama:
@lamminade looks good to me, thank you

Issue #113, opened 2021-03-05
Quick fix for Jakarta Bean Validation Constraint diagnostics


lamminade:
See: #53, ~~#63~~

> Add diagnostics for these constraints to ensure they are annotated on the supported types.

Quick fix should remove annotations on unsupported types

lamminade:
@kathrynkodama could you verify this is the desired behaviour for the quickfix and assign to me :)

kathrynkodama:
Sounds good to me, thanks @lamminade I would hold off on #63 since this is not yet implemented. Just focus on #53

PR #122, opened 2021-03-11
BeanValidation diagnostic collector extension + Quick Fixes

lamminade:
Combined the fixes for #113 and #114 because they are very similar:
- (113) if a field or method's return type are not compatible with the constraint annotation the quick fix should suggest removing the constraint
- (114) if a field or method with a constraint annotation have the `static` modifier, the quick fix should suggest either removing the constraint OR removing the `static` modifier

During implementation I determined that the constraint validation was not implemented for methods so I altered the diagnostic logic slightly to do so.

@kathrynkodama and I also determined that it made more sense for the warning to appear on the object name instead of the annotation as well.

Includes tests for both the diagnostic and quickfixes.

Screen Shot 2021-03-14 at 1 59 17 PM


kathrynkodama:
Edge case, when there are multiple invalid constraint annotations, only one diagnostic appears but two quick-fixes appear. Selecting either one will remove both of the invalid constraint annotation. Only one quick-fix should appear. 
image

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
```diff
@@ -0,0 +1,93 @@
+/*******************************************************************************
+* Copyright (c) 2020 IBM Corporation and others.
```
kathrynkodama, line 2: Header should be 2021

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
```diff
@@ -0,0 +1,93 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.IBinding;
+import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
+import org.eclipse.jdt.internal.corext.dom.Bindings;
+import org.eclipse.lsp4j.CodeAction;
+import org.eclipse.lsp4j.Diagnostic;
+import org.jakarta.codeAction.IJavaCodeActionParticipant;
+import org.jakarta.codeAction.JavaCodeActionContext;
+import org.jakarta.codeAction.proposal.AddConstructorProposal;
```
kathrynkodama, line 30: Remove unused imports

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
```diff
@@ -0,0 +1,93 @@
...
+            IProgressMonitor monitor) throws CoreException {
+        ASTNode node = context.getCoveredNode();
+        IBinding parentType = getBinding(node);
+
+        List<CodeAction> codeActions = new ArrayList<>();
+        codeActions.add(removeConstraintAnnotations(diagnostic, context, parentType));
+
+        if (diagnostic.getCode().getLeft().equals(BeanValidationConstants.DIAGNOSTIC_CODE_STATIC)) {
+            codeActions.add(removeStaticModifier(diagnostic, context, parentType));
+        }
```
kathrynkodama, line 57: When both static modifier and incorrect annotation is on the field, there is a duplicate code action:
image

kathrynkodama: Option to remove static modifier does not seem to appear on methods:
image

jakarta-eclipse/org.eclipse.lsp4jakarta.tests/src/main/java/org/eclipse/lsp4jakarta/jdt/beanvalidation/BeanValidationTest.java
```diff
@@ -0,0 +1,208 @@
+package org.eclipse.lsp4jakarta.jdt.beanvalidation;
```
kathrynkodama, line 1: Add header to this class

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
kathrynkodama: Consider adding the above scenarios to tests as well

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
lamminade: dcccd8d

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
lamminade: dcccd8d

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/beanvalidation/BeanValidationQuickFix.java
lamminade: dcccd8d

jakarta-eclipse/org.eclipse.lsp4jakarta.tests/src/main/java/org/eclipse/lsp4jakarta/jdt/beanvalidation/BeanValidationTest.java
lamminade: dcccd8d

kathrynkodama:
When there are multiple invalid constraint annotations, there are multiple diagnostic messages that appear:
image

Each of these diagnostics has an associated quick-fix, however it is confusing to see two identical quick-fixes. Additionally, selecting just one of these quick-fixes removes both of the invalid annotations. The expected behaviour should either be only one quick-fix appears or the associated quick-fix only removes the annotation the diagnostic refers to (the more ideal solution in my opinion).  It would be nice if the quick-fix message could also note which annotation(s) it would be removing. 
image

lamminade:
> When there are multiple invalid constraint annotations, there are multiple diagnostic messages that appear:
> 
> Each of these diagnostics has an associated quick-fix, however it is confusing to see two identical quick-fixes. Additionally, selecting just one of these quick-fixes removes both of the invalid annotations. The expected behaviour should either be only one quick-fix appears or the associated quick-fix only removes the annotation the diagnostic refers to (the more ideal solution in my opinion). It would be nice if the quick-fix message could also note which annotation(s) it would be removing.

fixed in a68d183

lamminade:
Not really sure why this is failing @kathrynkodama 😬

Issue #103, opened 2021-03-02
Jakarta CDI Scope diagnostic


Mandiiiiiii:
Add a diagnostic to validate that a bean class or producer method or field may specify at most one scope type annotation 
(https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#declaring_bean_scope)

So when we detected any 2 of `@RequestScoped`, `@ApplicationScoped`, `@SessionScoped`, `@ConversationScoped`, and `@Depenedent` are used for the same bean class/producer method/field, it should show a warning.

PR #126, opened 2021-03-24
Declaring bean scope

hanidamlaj:
Fix issue #103 

From https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#declaring_bean_scope:
"A bean class or producer method or field may specify at most one scope type annotation. If a bean class or producer method or field specifies multiple scope type annotations, the container automatically detects the problem and treats it as a definition error."

## Diagnostic
diagnostic

## Test using ```assertDiagnostics(List<Diagnostic> actual, List<Diagnostic> expected)```
- Ordering of the diagnostic within the actual and expected lists shouldn't affect equality – added support for ordering the two lists by the start and end ranges before comparing for equality

previously, assertDiagnostics would report the following error for two lists of identical diagnostics (fields of the diagnostic for brevity) due to the ordering within the list:

```
Failures:
  ManagedBeanTest.producesAndInject:70 Unexpected diagnostics:
expected:<[Diagnostic [
  range = Range [
    start = Position [
      line = 16
      character = 18
    ]
    end = Position [
      line = 16
      character = 23
    ]
  ]
], Diagnostic [
  range = Range [
    start = Position [
      line = 11
      character = 19
    ]
    end = Position [
      line = 11
      character = 27
    ]
  ]
]]> but was:<[Diagnostic [
  range = Range [
    start = Position [
      line = 11
      character = 19
    ]
    end = Position [
      line = 11
      character = 27
    ]
  ]
], Diagnostic [
  range = Range [
    start = Position [
      line = 16
      character = 18
    ]
    end = Position [
      line = 16
      character = 23
    ]
  ]
]]>
```

### DIFF
--- a/jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/cdi/ManagedBeanDiagnosticsCollector.java
+++ b/jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/cdi/ManagedBeanDiagnosticsCollector.java
@@ -102,6 +102,50 @@ public class ManagedBeanDiagnosticsCollector implements DiagnosticsCollector {
                                 DIAGNOSTIC_CODE);
                         diagnostics.add(diagnostic);
                     }
+
+                    /**
+                     * A bean class or producer method or field may specify at most one scope type
+                     * annotation. If a bean class or producer method or field specifies multiple
+                     * scope type annotations, the container automatically detects the problem and
+                     * treats it as a definition error.
+                     * 
+                     * Here we only look at the fields.
+                     */
+                    List<IAnnotation> fieldAnnotations = Arrays.asList(field.getAnnotations());
+                    List<String> fieldScopes = getManagedBeanAnnotations(field);
+
+                    boolean isProducerField = fieldAnnotations.stream()
+                            .anyMatch(annotation -> annotation.getElementName().equals(ManagedBeanConstants.PRODUCES));
+
+                    if (isProducerField && fieldScopes.size() > 1) {
+                        diagnostics.add(createDiagnostic(field, unit,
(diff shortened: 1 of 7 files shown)


jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/cdi/ManagedBeanDiagnosticsCollector.java
```diff
@@ -102,54 +102,104 @@ public void collectDiagnostics(ICompilationUnit unit, List<Diagnostic> diagnosti
...
+
+                    boolean isProducerMethod = methodAnnotations.stream()
+                            .anyMatch(annotation -> annotation.getElementName().equals(ManagedBeanConstants.PRODUCES));
+
+                    boolean isInjectMethod = methodAnnotations.stream()
+                            .anyMatch(annotation -> annotation.getElementName().equals(ManagedBeanConstants.INJECT));
+
+                    if (isProducerMethod && methodScopes.size() > 1) {
+                        diagnostics.add(createDiagnostic(method, unit,
+                                "A bean class or producer method or field may specify at most one scope type annotation.",
```
kathrynkodama, line 174: Tailor the diagnostic message to clarify if it is a field/producer method/ bean class

kathrynkodama (commented):
From the spec it sounds like there are 3 cases here:
1) A bean class with multiple scopes
2) Producer method with mulitple scopes
3) Field with multiple scopes

From what I see this is only covering case 2 but please correct me if I am wrong :). Can we expand to cover all 3 cases?

kathrynkodama (approved):
Looks good to me, thank you @hanidamlaj

Issue #119, opened 2021-03-05
Diagnostic for Root Resource Class with Multiple Public Constructors with Equal Number of Parameters


MShocrylas:
From [3.1.2](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#resource_class_constructor):
> Choosing amongst suitable constructors with the same number of parameters is implementation specific, implementations SHOULD generate a warning about such ambiguity.

A root resource class is: A resource class annotated with @​Path.

If a root resource class has multiple public constructors with the same number of parameters, create a warning that the chosen constructor is implementation specific and may be ambiguous.

See related #118

kathrynkodama:
Again we may not need a quick-fix for this issue

MShocrylas:
Removed quick fix from here as well

Issue #118, opened 2021-03-05
Diagnostic for Root Resource Class with Multiple Public Constructors of Different Length


MShocrylas:
From [3.1.2](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#resource_class_constructor):
> If more than one public constructor is suitable then an implementation MUST use the one with the most parameters.

A root resource class is: A resource class annotated with @​Path.

If a root resource class has multiple public constructors with different numbers of parameters, create a warning that the ones with less parameters are not used.

See related #119

kathrynkodama:
I think we can display a warning for this issue, but we may not need a quick-fix for it as developers may want to keep the other constructors

kathrynkodama:
Perhaps this could be tagged `DiagnosticSeverity.Warning` as opposed to Error

MShocrylas:
Removed the quick fix

PR #128, opened 2021-03-25
118/119 Root Resource Class Multiple Constructors

MShocrylas:
Implemented diagnostics for issues #118 and #119


MShocrylas:
We should check if #119 should be labeled as a warning or error.

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/jax_rs/RootResourceClassDiagnosticsCollector.java
```diff
@@ -0,0 +1,120 @@
...
+                                diagnostics.add(diagnostic);
+                            }
+                        }
+                        if (equalMaxParamMethods.size() > 1) {
+                            for (IMethod method : equalMaxParamMethods) {
+                                ISourceRange methodNameRange = JDTUtils.getNameRange(method); // TODO: maybe change the area this diagnostic underlines
+                                Range methodRange = JDTUtils.toRange(unit, methodNameRange.getOffset(), methodNameRange.getLength());
+                                
+                                diagnostic = new Diagnostic(methodRange, "Multiple constructors have the same number of parameters, it may be ambiguous which constructor is used.");
+                                diagnostic.setSeverity(Jax_RSConstants.SEVERITY_ERROR);
```
kathrynkodama, line 107: I think this should be a warning as well, to stay consistent and since it's not a breaking error. Let me know what you think @MShocrylas and @berasogut

kathrynkodama: If we change this to a warning we can also remove explicitly setting the severities here and instead just call `completeDiagnostic` since you set the severity there as well.

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/jax_rs/Jax_RSConstants.java
```diff
@@ -30,7 +30,8 @@
     public static final String DIAGNOSTIC_SOURCE = "jakarta-jax_rs";
 
     /* Severity */
-    public static final DiagnosticSeverity SEVERITY = DiagnosticSeverity.Error;
+    public static final DiagnosticSeverity SEVERITY_ERROR = DiagnosticSeverity.Error;
+    public static final DiagnosticSeverity SEVERITY_WARNING = DiagnosticSeverity.Warning;
```
kathrynkodama, line 34: Feels redundant to set these as constants in every constants class. Can we remove these from the constants class and instead just call `DiagnosticSeverity.Warning/Error` directly in the `completeDiagnostic` method below.

There are a number of other constants classes that do the same thing that we should fix.

jakarta-eclipse/org.eclipse.lsp4jakarta.core/src/org/jakarta/jdt/jax_rs/RootResourceClassDiagnosticsCollector.java
```diff
@@ -0,0 +1,120 @@
...
+                            }
+                        }
+                        if (equalMaxParamMethods.size() > 1) {
+                            for (IMethod method : equalMaxParamMethods) {
+                                ISourceRange methodNameRange = JDTUtils.getNameRange(method); // TODO: maybe change the area this diagnostic underlines
+                                Range methodRange = JDTUtils.toRange(unit, methodNameRange.getOffset(), methodNameRange.getLength());
+                                
+                                diagnostic = new Diagnostic(methodRange, "Multiple constructors have the same number of parameters, it may be ambiguous which constructor is used.");
+                                completeDiagnostic(diagnostic);
+                                diagnostics.add(diagnostic);
```
kathrynkodama, line 109: We should add an automated test for these new diagnostics

kathrynkodama (approved):
Looks good to me, thank you @MShocrylas

PR #141, opened 2021-04-27
Quickfix for ManagedBeanScopeDecl

hanidamlaj:
Quickfix for #126 – we utilize context Object to report the dynamic annotations the diagnostic may have.

From https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#declaring_bean_scope:
"A bean class or producer method or field may specify at most one scope type annotation. If a bean class or producer method or field specifies multiple scope type annotations, the container automatically detects the problem and treats it as a definition error."

- A bean class specifying more than one scope type annotation
preclass_quickfix

- Quick-fix for class specifying more than one scope type annotation
postclass_quickfix

- A producer field specifying more than one scope type annotation
prefield_quickfix

- Quick-fix for producer method specifying more than one scope type annotation
postfield_quickfix

Known issue – when the quick-fix removes the last annotation, it will remove the new line as well causing two lines to be merged. 

e.g. removing `@Dependent` scope from `private int n` results in the following:
bug

Signed-off-by: hanidamlaj <hanidamlaj@gmail.com>


jakarta-eclipse/org.eclipse.lsp4jakarta.tests/src/main/java/org/eclipse/lsp4jakarta/jdt/cdi/ManagedBeanTest.java
```diff
@@ -59,12 +59,15 @@ public void scopeDeclaration() throws Exception {
...
 
         Diagnostic d2 = d(15, 25, 41, "A producer method may specify at most one scope type annotation.",
                 DiagnosticSeverity.Error, "jakarta-cdi", "InvalidScopeDecl");
+        d2.setData(Arrays.asList("ApplicationScoped", "RequestScoped", "Produces"));
         
         Diagnostic d3 = d(10, 13, 29, "A managed bean class may specify at most one scope type annotation.",
                 DiagnosticSeverity.Error, "jakarta-cdi", "InvalidScopeDecl");
+        d3.setData(Arrays.asList("ApplicationScoped", "RequestScoped"));
 
         assertJavaDiagnostics(diagnosticsParams, JDT_UTILS, d1, d2, d3);
```
kathrynkodama, line 72: Add an assert statement for the associated quick-fix

Issue #26, opened 2020-10-27
Diagnostics delayed by 1 keystroke 


rzgry:
Diagnostics added by https://github.com/MicroShed/jakarta-ls/pull/23 are always 1 keystroke behind.

kathrynkodama:
The compilation unit being resolved here: https://github.com/MicroShed/jakarta-ls/blob/master/jakarta-eclipse/src/org/jakarta/jdt/JDTUtils.java#L111 is one keystroke behind the current document.

This results in an empty diagnostics list:  https://github.com/MicroShed/jakarta-ls/blob/master/jakarta-eclipse/src/org/jakarta/jdt/JDTServicesManager.java#L51-L53

PENGYUXIONG:
I dont have much progress on this bug still. 

My thought is that, either 
https://github.com/MicroShed/jakarta-ls/blob/26641efb51a3fbd0e2f05118d0d4f1548395418d/jakarta-eclipse/src/org/jakarta/jdt/JDTUtils.java#L117

is returning incorrect Ifile which I am not sure how to verify

or it can possibly be `JavaCore.createCompilationUnitFrom(resource);` from
https://github.com/MicroShed/jakarta-ls/blob/26641efb51a3fbd0e2f05118d0d4f1548395418d/jakarta-eclipse/src/org/jakarta/jdt/JDTUtils.java#L125  is not behave properly

kathrynkodama:
Adding the following to https://github.com/MicroShed/jakarta-ls/blob/master/jakarta.ls/src/main/java/io/microshed/jakartals/JakartaLanguageServer.java#L41 will update diagnostics on save: 
```
serverCapabilities.setTextDocumentSync(TextDocumentSyncKind.Incremental);
```
Keystroke delay still exists without saving

tiganov:
Adding a delay before `createCompilationUnitFrom` is called [in JDTUtils.java](https://github.com/eclipse/lsp4jakarta/blob/26641efb51a3fbd0e2f05118d0d4f1548395418d/jakarta-eclipse/src/org/jakarta/jdt/JDTUtils.java#L125) seems to resolve the keystroke delay, but I'm not quite sure why, yet.

```java
try {
    Thread.sleep(1000);
} catch (InterruptedException e) {
    e.printStackTrace();
}
return JavaCore.createCompilationUnitFrom(resource);
```

kathrynkodama:
Closed with #201

PR #201, opened 2021-10-29
Fixed keystroke delay (#26)

tiganov:
See my [comment](https://github.com/eclipse/lsp4jakarta/issues/26#issuecomment-954398977) in issue #26.

I decided to put the loop in `JDTUtils.resolveCompilationUnit()` so that anything that uses it has a consistent `ICompilationUnit`. I set the while loop timeout to 3 seconds, just in case.


kathrynkodama (approved):
LGTM, thanks @tiganov!

PR #213, opened 2021-11-16
Added diagnostics and quickfix for a managed bean class of generic type for #210

### LINKED ISSUE #210: Dependency Injection: Diagnostics for bean class of generic or parameterized type
[Managed Beans](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#managed_beans)

> If the managed bean class is a generic type, it must have scope @Dependent. If a managed bean with a parameterized bean class declares any scope other than @Dependent, the container automatically detects the problem and treats it as a definition error.

## Diagnostics
1. Check if class is a [managed bean](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#what_classes_are_beans)
2. Check if managed bean is of generic/parameterized type
3. Check if dependency class is annotated with `@Dependent` or has no other scope specified
4. If not, deliver an error at the class declaration

## Quickfixes
- Suggest removing scope

## Example
```java
@ApplicationScoped
// Error here due to the generic type
public class Greeting<T>{
	public String greet(String name) {
        return "GreetingA, " + name;
    }
}
```

```
[INFO] [ERROR   ] CWWKZ0004E: An exception occurred while starting the application demo-servlet. The exception message was: com.ibm.ws.container.service.state.StateChangeException: org.jboss.weld.exceptions.DefinitionException: WELD-000071: Managed bean with a parameterized bean class must be @Dependent: class io.openliberty.sample.jakarta.di.error.generic.Greeting
```

Related to #153

AlvinTan2000:
Fixes #210 

Will implement integration test.


AlvinTan2000:
Closing PR because it does not check if the class is a managed bean.

Issue #159, opened 2021-09-18
WebServlet annotation type check is unsound


tiganov:
The [ServletDiagnosticsCollector.java](https://github.com/eclipse/lsp4jakarta/blob/f828839a94d9fac3a355905ee6d807861eae4d76/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/servlet/ServletDiagnosticsCollector.java#L69) checks if a class annotated with `@WebServlet` extends `HttpServlet`. However, if the user extends a class that extends `HttpServlet` instead of extending `HttpServlet` directly, they will still have a diagnostic. e.g.,

```
class A extends HttpServlet { ... }

@WebServlet
class B extends A { ... }
```

While rare, this does happen in reality. e.g., [Spring](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/HttpServletBean.html) and [oath2 client](https://github.com/eugenp/tutorials/blob/master/oauth2-framework-impl/oauth2-client/src/main/java/com/baeldung/oauth2/client/AbstractServlet.java).

We would need sound type hierarchy information to have proper support for this diagnostic. This is just one case where we need a type hierarchy. Another example: the [setter based injection](https://jakarta.ee/specifications/annotations/2.0/annotations-spec-2.0.html#setter-based-injection) of `jakarta.annotation.Resource`(#157) requires the `type` attribute of the annotation to be compatible with the return type of the method. In this case, we would also need type hierarchy information to be sound.

Naturally, we do not have all of the source code information to develop a complete type hierarchy. Even if we did, reasoning about different package structures and imports is difficult.

Suggestion: Lower diagnostic from error to warning level for `@WebServlet`.

kathrynkodama:
LSP4MP has an annotation for `mprestclient` that appears if the field on current type has `@Inject` and `@RestClient` annotations but the corresponding interface does not have `@RegisterRestClient` annotation. We can take a look at how they gather information that the corresponding interface does not have the `@RegisterRestClientAnnotation` and see if we can do something similar in this case.

https://github.com/eclipse/lsp4mp/blob/master/microprofile.jdt/org.eclipse.lsp4mp.jdt.core/src/main/java/org/eclipse/lsp4mp/jdt/internal/restclient/java/MicroProfileRestClientDiagnosticsParticipant.java

cc: @ericglau

Rejoice472:
[...]

PR #214, opened 2021-11-16
Added type hierarchy check for @WebServlet diagnostic (#159)

tiganov:
**DRAFT:** The tests for @WebServlet diagnostics fail because my fix will report a diagnostic that the WebServlet class does not extend `HttpServlet` when in fact it does. This is because under the test environment `typeHierarchy.getAllSupertypes(type);` is empty. I've tried the test code in my eclipse instance with my change and no diagnostic is reported.

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/TypeHierarchyUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/TypeHierarchyUtils.java
@@ -1,3 +1,16 @@
+/*******************************************************************************
+* Copyright (c) 2021 IBM Corporation and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     IBM Corporation - initial API and implementation
+*******************************************************************************/
+
 package org.eclipse.lsp4jakarta.jdt.core;
 
 import java.util.concurrent.atomic.AtomicInteger;
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/TypeHierarchyUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/TypeHierarchyUtils.java
@@ -1,0 +1,99 @@
+package org.eclipse.lsp4jakarta.jdt.core;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.ITypeHierarchy;
+import org.eclipse.jdt.core.search.IJavaSearchConstants;
+import org.eclipse.jdt.core.search.IJavaSearchScope;
+import org.eclipse.jdt.core.search.SearchEngine;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchParticipant;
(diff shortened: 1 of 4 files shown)


kathrynkodama:
Looking at the test failures:
```
expected:<[Diagnostic [
  range = Range [
    start = Position [
      line = 9
      character = 0
    ]
    end = Position [
      line = 9
      character = 13
    ]
  ]
  severity = Error
  code = Either [
    left = CompleteHttpServletAttributes
    right = null
  ]
  codeDescription = null
  source = "jakarta-servlet"
  message = "The 'urlPatterns' attribute or the 'value' attribute of the WebServlet annotation MUST be specified."
  tags = null
  relatedInformation = null
  data = null
]]> but was:<[Diagnostic [
  range = Range [
    start = Position [
      line = 10
      character = 13
    ]
    end = Position [
      line = 10
      character = 30
    ]
  ]
  severity = Error
  code = Either [
    left = ExtendHttpServlet
    right = null
  ]
  codeDescription = null
  source = "jakarta-servlet"
  message = "Classes annotated with @WebServlet must extend the HttpServlet class."
  tags = null
  relatedInformation = null
  data = null
], Diagnostic [
  range = Range [
    start = Position [
      line = 9
      character = 0
    ]
    end = Position [
      line = 9
      character = 13
    ]
  ]
  severity = Error
  code = Either [
    left = CompleteHttpServletAttributes
    right = null
  ]
  codeDescription = null
  source = "jakarta-servlet"
  message = "The 'urlPatterns' attribute or the 'value' attribute of the WebServlet annotation MUST be specified."
  tags = null
  relatedInformation = null
  data = null
]]>
	at org.eclipse.lsp4jakarta.jdt.servlet.JakartaServletTest.CompleteWebServletAnnotation(JakartaServletTest.java:73)

```

The [CompleteWebServletTest](https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/servlet/JakartaServletTest.java#L60) is failing as it detects two diagnostics but only expects 1. 

With your new solution, the `extends HttpServlet` diagnostic is being detected if it exists anywhere in the project, rather than just the open file. Since the files are in the same project: https://github.com/eclipse/lsp4jakarta/tree/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/servlet it is picking up the `extends HttpServlet` even though the problem does not exist in the currently open file.

This seems like an improvement and then we should adjust the `CompleteWebServletTest` to expect both diagnostics (or combine the two test cases)

tiganov:
For some reason `getAllSupertypes()` returns empty when it should return `HttpServlet` for `InvalidWebServlet`, but only when running that test. If I take the test file and run it in an Eclipse instance, no diagnostic is reported because `getAllSupertypes()` returns the correct values. I've disabled the tests as per @kathrynkodama's instructions. Once this PR goes through, I'll open a new issue for fixing this test problem.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/TypeHierarchyUtils.java
```diff
@@ -0,0 +1,99 @@
+package org.eclipse.lsp4jakarta.jdt.core;
```
kathrynkodama, line 1: This class needs a header, see an example: https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/jax_rs/NoResourcePublicConstructorQuickFix.java#L1-L12

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/servlet/JakartaServletTest.java
```diff
@@ -57,6 +59,7 @@ public void ExtendWebServlet() throws Exception {
     }
 
     @Test
+    @Ignore
```
kathrynkodama, line 62: Let's add a comment indicating why we are ignoring this test for now

kathrynkodama (approved):

PR #230, opened 2021-11-30
added Provider quick fix/diagnostic, renamed some files, added tests …

### LINKED ISSUE #124: Diagnostic and quick fix for Provider class constructors
From [4.1.2](https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#provider_class_constructor):

> Provider classes that are instantiated by the JAX-RS runtime MUST have a public constructor for which the JAX-RS runtime can provide all parameter values. Note that a zero argument constructor is permissible under this rule.

A provider class is a class annotated with @​Provider.

Create a warning if a provider class with at least 1 defined constructor has no public constructor. (If no constructors are defined, Java creates a default public constructor)

Quick fixes:
- Create empty public constructor
- Make non public constructor public (optional)

See similar: #116

Tested with different combinations of constructors in a provider class. Received this error with only a private constructor:
Screen Shot 2021-03-18 at 12 03 00 PM

shaunak:
Fixes #124

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
@@ -1,0 +1,16 @@
+package io.openliberty.sample.jakarta.jax_rs;
+
+import jakarta.ws.rs.ext.Provider;
+
+@Provider()
+public class NoPublicConstructorProviderClass {
+
+    private NoPublicConstructorClass() {
+
+    }
+
+    protected NoPublicConstructorClass(int arg1) {
+
+    }
+
+}
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
@@ -2,14 +2,16 @@ package io.openliberty.sample.jakarta.jax_rs;
 
 import jakarta.ws.rs.ext.Provider;
 
-@Provider()
-public class NoPublicConstructorProviderClass {
 
-    private NoPublicConstructorClass() {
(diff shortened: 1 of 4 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
```diff
@@ -0,0 +1,16 @@
+package io.openliberty.sample.jakarta.jax_rs;
+
+import jakarta.ws.rs.ext.Provider;
+
+@Provider()
+public class NoPublicConstructorProviderClass {
+
+    private NoPublicConstructorClass() {
```
kathrynkodama, line 8: Should be: 
```java
private NoPublicConstructorProviderClass() {
}
```

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
```diff
@@ -0,0 +1,16 @@
...
+
+    }
+
+    protected NoPublicConstructorClass(int arg1) {
```
kathrynkodama, line 12: Should be:
```java
protected  NoPublicConstructorProviderClass(int arg1){
}
```

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/NoPublicConstructorProviderClass.java
```diff
@@ -0,0 +1,16 @@
+package io.openliberty.sample.jakarta.jax_rs;
+
+import jakarta.ws.rs.ext.Provider;
+
+@Provider()
+public class NoPublicConstructorProviderClass {
```
kathrynkodama, line 6: Is this how a Provider class typically looks? We should try and model this test class as a simple class showcasing how `@Provider` is typically used in practice

See examples shown here: https://eclipse-ee4j.github.io/jakartaee-tutorial/#using-entity-providers-to-map-http-response-and-request-entity-bodies

More examples here (but note that these are using the old javax namespace instead of jakarta): https://www.logicbig.com/how-to/code-snippets/jcode-jax-rs-provider.html

kathrynkodama (approved):

PR #231, opened 2021-12-01
Rephrasing diagnostic messages.

### LINKED ISSUE #189: Rephrase diagnostic messages and edit ranges
ie. if an attribute is missing from an annotation, the diagnostic from jdt core is `The annotation @{0} must define the attribute {1}`.

We should modify our diagnostic messages to match similar phrasing.
For example, one of the servlet diagnostics is `The 'urlPatterns' attribute or the 'value' attribute of the WebServlet annotation MUST be specified.` but should be `The annotation @WebServlet must define the attribute urlPatterns or the attribute value`
 See https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/servlet/ServletDiagnosticsCollector.java#L105. 

This applies to similar diagnostics for `@WebServlet` and may apply to a few others. Review of the different diagnostics we have for annotations is required.

ananya2801:
Rephrased diagnostic messages to match similar phrasing as the jdt core.

Fixes #189.

### DIFF
--- a/.github/workflows/lsp4jakarta.yaml
+++ b/.github/workflows/lsp4jakarta.yaml
@@ -2,9 +2,9 @@ name: Java CI - LSP4Jakarta
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
-    branches: [master]
+    branches: [main]
 
 jobs:
   build:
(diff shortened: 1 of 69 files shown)


kathrynkodama (commented):
These messages look good @ananya2801, if we can make a common class for similar messages that would be even better

kathrynkodama (approved):
Thank you @ananya2801 looks good to me. You just have to mark "Ready for review" so that this is no longer a draft PR

PR #233, opened 2021-12-08
Update p2 repos to fix build failures

kathrynkodama:
Signed-off-by: Kathryn Kodama <kathryn.s.kodama@gmail.com>


jakarta.jdt/pom.xml
```diff
@@ -52,7 +52,7 @@
 						<artifact>
 							<groupId>org.eclipse.jdt.ls</groupId>
 							<artifactId>org.eclipse.jdt.ls.tp</artifactId>
-							<version>${jdt.ls.version}</version>
+							<version>1.7.0.20211216164144</version>
```
yeekangc, line 55: Why aren't we using the property here, @kathrynkodama?

kathrynkodama: We should be using the property, overlooked this thank you. Updated to utilize the property

yeekangc (approved):
LGTM.

PR #254, opened 2022-02-16
Created jsonp package and constants

jooyoojoo:
Created jsonp package. Added jsonp diagnostic collector and constants.


kathrynkodama (approved):
Looks good to me, thanks @jooyoojoo!

PR #258, opened 2022-02-23
Add diagnostic support for ServerEndpoint URI validation

### LINKED ISSUE #240: ServerEndpoint URI validation
#### Description:
Validate the strings for the `@ServerEndpoint` annotation and provide diagnostic information when a URI is not valid.

#### Specification:
All server endpoint paths must:
- be a URI-template (level-1) or a partial URI
- start with a leading `/`
- not contain the sequences `/../`, `/./` or `//`

[WebSocket Specification](https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html#uri-mapping)

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

giancarlopernudisegura:
Resolves #240 
Create methods to:
+ Verify URIs have a leading slash (done in #263)
+ Verify that a URI is following a valid URI template (level 1)
+ Verify that a URI is not using duplicate variables
+ Verify that URIs are not using relative paths

These first two methods are located in JDTUtils so that other packages can access them for ease of use.

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -625,4 +625,21 @@ public class JDTUtils {
     public static boolean hasLeadingSlash(String uriString) {
         return uriString.startsWith("/");
     }
+
+    /**
+     * Check if a URI follows a valid URI-template (level-1) specified by
+     * <a href="https://datatracker.ietf.org/doc/html/rfc6570">RFC 6570</a>.
+     *
+     * @param uriString
+     * @return boolean
+     */
+    public static boolean isValidLevel1URI(String uriString, boolean isPartial) {
+        // Regex is obtained from RFC 3986:
+        // https://www.rfc-editor.org/rfc/rfc3986#appendix-B
+        if (isPartial) {
+            return uriString.matches("[^?#]*");
+        } else {
+            return uriString.matches("^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?");
+        }
+    }
 }
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -87,7 +87,7 @@ public class JDTUtils {
     public static final String DEFAULT_PROJECT_NAME = "jdt.java-project";
 
     private static final int COMPILATION_UNIT_UPDATE_TIMEOUT = 3000;
-    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{\\w+\\})|(?:\\w+)))*\\/?";
+    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\})|(?:(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+)))*\\/?";
 
     /**
      * Given the uri returns a {@link ICompilationUnit}. May return null if it can
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -85,8 +85,9 @@ public class JDTUtils {
     private static Set<String> SILENCED_CODEGENS = Collections.singleton("lombok");
 
     public static final String DEFAULT_PROJECT_NAME = "jdt.java-project";
-    
+
     private static final int COMPILATION_UNIT_UPDATE_TIMEOUT = 3000;
+    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{\\w+\\})|(?:\\w+)))*\\/?";
 
     /**
      * Given the uri returns a {@link ICompilationUnit}. May return null if it can
@@ -633,13 +634,7 @@ public class JDTUtils {
      * @param uriString
      * @return boolean
      */
-    public static boolean isValidLevel1URI(String uriString, boolean isPartial) {
-        // Regex is obtained from RFC 3986:
-        // https://www.rfc-editor.org/rfc/rfc3986#appendix-B
-        if (isPartial) {
-            return uriString.matches("[^?#]*");
-        } else {
-            return uriString.matches("^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?");
-        }
+    public static boolean isValidLevel1URI(String uriString) {
+        return uriString.matches(LEVEL1_URI_REGEX);
     }
 }
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -87,7 +87,7 @@ public class JDTUtils {
     public static final String DEFAULT_PROJECT_NAME = "jdt.java-project";
 
     private static final int COMPILATION_UNIT_UPDATE_TIMEOUT = 3000;
-    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\})|(?:(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+)))*\\/?";
+    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{(\\w|-|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\})|(?:(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+)))*\\/?";
 
     /**
      * Given the uri returns a {@link ICompilationUnit}. May return null if it can
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -92,6 +92,7 @@ public class JDTUtils {
     public static final String DEFAULT_PROJECT_NAME = "jdt.java-project";
 
     private static final int COMPILATION_UNIT_UPDATE_TIMEOUT = 3000;
+    // Percent encoding obtained from: https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters
     private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{(\\w|-|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\})|(?:(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+)))*\\/?";
 
     /**
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -643,6 +643,7 @@ public class JDTUtils {
         return uriString.matches(LEVEL1_URI_REGEX);
     }
 
+    /**
      * Returns a list of all accessors (getter and setter) of the given field.
      * 
      * @param field
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
@@ -88,6 +88,8 @@ public class WebSocketConstants {
     public static final String PARAM_TYPE_DIAG_MSG = "Invalid parameter type. When using %s, parameter must be of type: \n- %s\n- annotated with @PathParams and of type String or any Java primitive type or boxed version thereof";
 
     /* Regex */
+    // Check for any URI strings that contain //, /./, or /../
     public static final String REGEX_RELATIVE_PATHS = ".*\\/\\.{0,2}\\/.*";
+    // Check that a URI string is a valid level 1 variable (wrapped in curly brackets): alpha-numeric characters, dash, or a percent encoded character
     public static final String REGEX_URI_VARIABLE = "\\{(\\w|-|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\}";
 }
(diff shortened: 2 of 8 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
```diff
@@ -43,19 +43,25 @@
...
     public static final String DIAGNOSTIC_CODE_PATH_PARMS_ANNOT = "AddPathParamsAnnotation";
-    
+
     public static final String DIAGNOSTIC_ON_OPEN_INVALID_PARAMS = "Invalid parameter type. Parameter must be of type: \n- jakarta.websocket.Session \n- jakarta.websocket.EndpointConfig \n- annotated with @PathParams and of type String or any Java primitive type or boxed version thereof";
     public static final String DIAGNOSTIC_CODE_ON_OPEN_INVALID_PARAMS = "ChangeInvalidParam";
-    
+
+    public static final String DIAGNOSTIC_SERVER_ENDPOINT_NO_SLASH = "Server endpoint paths must start with a leading '/'.";
+    public static final String DIAGNOSTIC_SERVER_ENDPOINT_NOT_LEVEL1 = "Server endpoint paths must be a URI-template (level-1) or a partial URI.";
+    public static final String DIAGNOSTIC_SERVER_ENDPOINT_RELATIVE = "Server endpoint paths must not contain the sequences /../, /./ or //.";
```
TheAvi123, line 52: for consistency and clarity, it would be nice to place the actual sequences within single quotes like the diagnostic message on line 50. Specifically, the last one might be confusing and a user might think the sequence is '//.' rather than just '//'

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -92,6 +93,7 @@ public void collectDiagnostics(ICompilationUnit unit, List<Diagnostic> diagnosti
                             WebSocketConstants.RAW_ON_OPEN_PARAM_OPT_TYPES, unit, diagnostics);
                     // PathParam URI Mismatch Warning Diagnostic
                     uriMismatchWarningCheck(type, diagnostics, unit);
+                    serverEndpointWarningCheck(type, diagnostics, unit);
```
TheAvi123, line 96: again, just for clarity, if you could add a comment above this call to separate this diagnostic from the URI mismatch diagnostic, it would be helpful

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -310,4 +348,35 @@ private boolean isWrapper(String valueClass) {
 
         return wsEndpoint;
     }
+
+    /**
+     * Check if a URI string contains any sequence with //, /./, or /../
+     *
+     * @param uriString
+     * @return
```
TheAvi123, line 356: specification comment doesn't seem to be consistent with the method signature

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -310,4 +348,35 @@ private boolean isWrapper(String valueClass) {
...
+     */
+    private boolean hasRelativePathURIs(String uriString) {
+        return uriString.matches(WebSocketConstants.REGEX_RELATIVE_PATHS);
+    }
+
+    /*
+     * Check if a URI string has a duplicate variable
+     * 
+     * @param uriString
+     * @return
```
TheAvi123, line 366: fix comment

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -202,6 +204,42 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
     }
 
+    /*
+     * Create a warning diagnostic if a ServerEndpoint annotation does not follow the rules.
+     */
+    private void serverEndpointWarningCheck(IType type, List<Diagnostic> diagnostics, ICompilationUnit unit) throws JavaModelException {
+        for (IAnnotation annotation : type.getAnnotations()) {
+            if (annotation.getElementName().equals(WebSocketConstants.SERVER_ENDPOINT_ANNOTATION)) {
+                String path = annotation
+                        .getMemberValuePairs()[0]
```
TheAvi123, line 214: Are you sure that the URI will always be the first memberValuePair present for the Server Endpoint annotation? The annotation has other possible attributes such as encoders/decoders, so if those are specified first this will cause some unexpected errors. Probably best to iterate through these pairs and explicitly check that the key for the memberValuePair is 'value'.

See https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html#serverendpoint for details.

TheAvi123 (changes requested):
Looks good for the most part. Highlighted a couple of minor changes for clarity/consistency and one potentially significant problem with getting the URI string from the endpoint annotation.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
giancarlopernudisegura: oh that's a good point

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: What do you mean? Checking if a URI has relative paths involves checking for `//`, `/./`, and `/../` sequences.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: fix in regards to what?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: I'll look into it.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -202,6 +193,45 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
         }
     }
 
+    /**
+     * Create a warning diagnostic if a ServerEndpoint annotation does not follow the rules.
```
dalidia, line 197: You are not creating a warning diagnostic. You are creating an error diagnostic. Did you mean to do it only a warning? If not, could you update the comments and the respective methods/variables that refer to it as a warning as this could lead to a confusion?

dalidia (changes requested):
Seems good but left a comment about the kind of diagnostic you are using and how you refer to it in the code

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: It was a typo, meant to be an error.

dalidia (approved):

TheAvi123 (approved):

kathrynkodama:
@turkeylurkey please review

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -204,6 +198,45 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+                        String path = annotationMemberValuePair
+                                .getValue()
+                                .toString();
+                        Diagnostic diagnostic;
+                        if (!JDTUtils.hasLeadingSlash(path)) {
+                            diagnostic = createDiagnostic(annotation, unit,
+                                    WebSocketConstants.DIAGNOSTIC_SERVER_ENDPOINT_NO_SLASH,
+                                    WebSocketConstants.DIAGNOSTIC_SERVER_ENDPOINT);
+                            diagnostics.add(diagnostic);
+                        } else if (hasRelativePathURIs(path)) {
```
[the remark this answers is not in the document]
giancarlopernudisegura: I originally had it as simple ifs but with my testing I never got more than one diagnostic to show up so I ended up switching it to else if. I have no problem switching it back maybe there's an edge case I didn't try (like the one you pointed out).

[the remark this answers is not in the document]
giancarlopernudisegura: Just tried the new changes with your test case and some variations for it. I still don't get more than one diagnostic to show up at a time and any `//`, `/./`, or `/../` sequences don't trigger their respective diagnostic message but show the URI template one instead. Do you want me and go ahead with the new changes?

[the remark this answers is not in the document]
turkeylurkey: When I ran it I saw strange exceptions in the console from org.eclipse.lsp4e, the framework. When I stepped through your code in the debugger it was working properly and I did see all the messages in the UI:
image
I think the failures you are seeing could be caused by bugs in the framework. 
I find the relative path URI message appears but that the Level 1 URI message also always appears. That makes sense because `//` is going to fail both checks. For that reason it makes sense to test the relative path URI first because it is more specific and use `else if` before the Level 1 URI check so it is only run if the relative path check is ok. The other checks don't interfere with each other. 
It is usually best to flag all the errors at once so the person can fix them all at once. I think your code would be more correct using `else if` just in the one case. 
You could investigate issues against LSP4E to see if any of them match this. Possibly open an issue if you could make a simple test case that demonstrates it. Even investigate the cause.
@kathrynkodama  what do you think? Just generate one message to work around a possible issue in the framework or generate multiple (if required) and hope the framework gets fixed in the future?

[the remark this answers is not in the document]
kathrynkodama: If I am understanding the above correctly, there are currently multiple diagnostics being shown on the same line (the multiple markers in that screenshot above). 
I would think that only the most relevant diagnostic message is displayed to the user.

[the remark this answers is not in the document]
giancarlopernudisegura: So keep it at one diagnostic message max with `else if`?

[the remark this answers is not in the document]
kathrynkodama: Re-reading the above again and I may have misunderstood. The multiple diagnostics appear when there is more than one relevant diagnostic message correct? In that case, I agree with Paul:
> It is usually best to flag all the errors at once so the person can fix them all at once. I think your code would be more correct using else if just in the one case.
This makes sense the most sense to me. If there are issues in the framework when implementing this solution let's open a follow-up bug.

turkeylurkey (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/websocket/ServerEndpointDuplicateVariableURI.java
```diff
@@ -0,0 +1,6 @@
+package src.main.java.io.openliberty.sample.jakarta.websocket;
+
+import jakarta.websocket.server.ServerEndpoint;
+
+@ServerEndpoint("/{varA}/{varB}/{varA}")
```
kathrynkodama, line 5: can we add some comments to these test classes indicating where the diagnostic/error is expected and why? Can be brief comments

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -204,6 +198,47 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+                    if (annotationMemberValuePair.getMemberName().equals(WebSocketConstants.ANNOTATION_VALUE)) {
+                        String path = annotationMemberValuePair
+                                .getValue()
+                                .toString();
+                        Diagnostic diagnostic;
+                        if (!JDTUtils.hasLeadingSlash(path)) {
+                            diagnostic = createDiagnostic(annotation, unit,
+                                    WebSocketConstants.DIAGNOSTIC_SERVER_ENDPOINT_NO_SLASH,
+                                    WebSocketConstants.DIAGNOSTIC_SERVER_ENDPOINT);
+                            diagnostics.add(diagnostic);
```
kathrynkodama, line 217: This diagnostic seems like it would be a good candidate for a quick fix, to prefix the string with a leading '/'. What do you think @giancarlopernudisegura? If you agree let's create a follow-up issue if we do not have one already

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
```diff
@@ -90,8 +90,9 @@
     private static Set<String> SILENCED_CODEGENS = Collections.singleton("lombok");
 
     public static final String DEFAULT_PROJECT_NAME = "jdt.java-project";
-    
+
     private static final int COMPILATION_UNIT_UPDATE_TIMEOUT = 3000;
+    private static final String LEVEL1_URI_REGEX = "(?:\\/(?:(?:\\{(\\w|-|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\})|(?:(\\w|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+)))*\\/?";
```
kathrynkodama, line 95: Was this referenced from somewhere?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
```diff
@@ -81,4 +86,8 @@
 
     // Messages
     public static final String PARAM_TYPE_DIAG_MSG = "Invalid parameter type. When using %s, parameter must be of type: \n- %s\n- annotated with @PathParams and of type String or any Java primitive type or boxed version thereof";
+
+    /* Regex */
+    public static final String REGEX_RELATIVE_PATHS = ".*\\/\\.{0,2}\\/.*";
+    public static final String REGEX_URI_VARIABLE = "\\{(\\w|-|%20|%21|%23|%24|%25|%26|%27|%28|%29|%2A|%2B|%2C|%2F|%3A|%3B|%3D|%3F|%40|%5B|%5D)+\\}";
```
kathrynkodama, line 92: Consider some (small) detailing comments explaining each regex

kathrynkodama:
Looks good to me, thanks @giancarlopernudisegura just some minor comments/questions above

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
giancarlopernudisegura: I got the list from here: https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: I agree this would be a good quick fix. I will create a follow-up issue as there isn't one already.

giancarlopernudisegura: Created the issue: #284

kathrynkodama (approved):
Looks good to me, thanks @giancarlopernudisegura!

PR #260, opened 2022-02-24
On open params validation

### LINKED ISSUE #247: WebSocket: Validate for parameters OnOpen lifecycle method
#### Description:
Currently, if a user adds a non-valid parameter to methods annotated with `@OnOpen` to classes annotated with `@ServerEndpoint` or `@ClientEndpoint`, it will create a server error, leaving the server unresponsive. However, the user is not notified of this in the IDE. Only a small error shows up in the terminal.

For this reason, we need to apply a validation of parameters for the `OnOpen` lifecycle method.

#### Specification:
Valid parameters are defined as "an optional Session parameter, an optional EndpointConfig parameter, and zero to n parameters (of type String, any Java primitive type or boxed version thereof) annotated with a @PathParam annotation as parameters" per JakartaEE specs found [here](https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html#onopen)

##### Acceptable
```java
@OnOpen
public void validOpenHandler(EndpointConfig endpointConfig, Session session) {
    System.out.println("Program requested " + message + " using " + session.getId());
    session.getAsyncRemote().sendText("Lucky says hi");
}
```

##### Error
```java
@OnOpen
public void notValidOpenHandler(EndpointConfig endpointConfig, String invalidParameter) {
   System.out.println("Program requested " + invalidParameter + " using " + session.getId());
   session.getAsyncRemote().sendText("I am invalid");
}
```

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

dalidia:
- Check that a method annotated with `OnOpen` located in classes annotated with `@ServerEndpoint` or `@ClientEndpoint` have valid parameters, that is, "an optional Session parameter, an optional EndpointConfig parameter, and zero to n parameters (of type String, any Java primitive type or boxed version thereof) annotated with a @PathParam annotation as parameters". However, currently it's not checking if the type for parameters annotated with  `@PathParam`
- Resolves #247 
- Opened an issue #265 for the test projects signature types not resolving correctly


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -13,25 +13,181 @@
...
+		// check methods annotations, then check their params, check that their type is
+		// any of the valid options, if it's a string, check that it has the @Params
+		// annotation
+		// If so, set up the severity to Error, and add a code?
+
+		IMethod[] allMethods = type.getMethods();
+
+		for (IMethod method : allMethods) {
+			IAnnotation[] allAnnotations = method.getAnnotations();
+
```
kathrynkodama, line 113: On idea for getting the types on the method params: 

Example:
```java
@OnOpen
public void OnOpen(Integer boxIntParam, int intParam, String stringParam) throws IOException {
```

Iterating through the method.getParameters:
```java
for (ILocalVariable var : method.getParameters()) {
	String elemName = var.getElementName(); // resolves to boxIntParam, intParam, stringParam
	String typeSignature = var.getTypeSignature(); // resolves to QInteger, I, QString
}
```

Here is some info on type signatures: https://www.ibm.com/docs/en/rational-soft-arch/9.6.1?topic=SS8PJ7_9.6.1/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/core/Signature.html 
Looks like Integer and String are "UnresolvedClassTypeSignatures"... so thinking we may need to hardcode check for these values.

kathrynkodama: Improved on the above slightly:

```java

for (ILocalVariable var : method.getParameters()) {
	String elemName = var.getElementName(); // resolves to boxIntParam, intParam, stringParam
	String typeSignature = var.getTypeSignature(); // resolves to QInteger, I, QString
	String signature = var.getTypeSignature().replace("/", ".");
	 IType primaryType = var.getTypeRoot().findPrimaryType();
	String resolvedTypeName = JavaModelUtil.getResolvedTypeName(signature, primaryType); // resolved to java.lang.Integer, int, java.lang.String
	boolean test = true;

}

```

kathrynkodama: @dalidia  This seems like the best way I can find so far, then we may still need to hardcode the accepted primitive and boxed primitive types and open a follow up issue to improve upon this. Let me know what you think
cc: @turkeylurkey @TrevCraw

dalidia: Thanks Kathryn! I ended up using that approach. Even though it works in manual testing, `JavaModelUtil.getResolvedTypeName(signature, primaryType)` fails to get the resolved type of a signature when writing the integration tests for all parameters type, returning a `null` instead. I debugged the tests and it successfully identifies the `primaryType`. However, it doesn't seem to be able to connect it to the signature. 

As an example of the current test behavior, when the integration tests run, the pluggin will always throw an error to the first parameter declared inside the method. This is because `resolvedTypeName` is [used to evaluate if it's a wrapper and if it's a Session](https://github.com/dalidia/lsp4jakarta/blob/onOpen-params-validation/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java#L116) In the example below, it will be session.
```java
@OnOpen
public void OnOpen(Session session, int intParam, String stringParam) throws IOException {
```

 @kathrynkodama  do you think the issue of not being able to get the resolved type is coming from the way the project was loaded for testing purposes? I was thinking that I could add a null checker for `JavaModelUtil.getResolvedTypeName(signature, primaryType)` and if it's null, either don't run diagnostics, or use the raw signature result (from `getTypeSignature`) . See example below:

```java
String signature = param.getTypeSignature();
String formatSignature = signature.replace("/", ".");
IType primaryType = param.getTypeRoot().findPrimaryType();
String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, primaryType);

if (resolvedTypeName == null) {
    // compare signature to "QString" || "QInteger" || ... instead of "java.lang.String" || "java.lang.Integer"
}
```

kathrynkodama: This is interesting, I like the second suggestion of adding a null check and using the raw signature if it is null, we should open a follow up issue (labelled as a bug) for the test projects signature types not resolving correctly

kathrynkodama:
@giancarlopernudisegura or @TheAvi123 can either of you please peer-review?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/AnnotationUtil.java
```diff
@@ -0,0 +1,31 @@
+package org.eclipse.lsp4jakarta.jdt.core;
```
TheAvi123, line 1: Since this is a new file, I think we need a license on top. I see you have one in your test class, so you should be able to just copy that into here.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -13,25 +13,196 @@
...
+					for (ILocalVariable param : allParams) {
+					    
+					    String signature = param.getTypeSignature();
+					    String formatSignature = signature.replace("/", ".");
+					    IType primaryType = param.getTypeRoot().findPrimaryType();
+					    String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, primaryType);
+					    					    
+					    boolean isPrimitive = JavaModelUtil.isPrimitive(formatSignature);
+					    boolean isSpecialType = specialParamTypes.contains(resolvedTypeName);
```
TheAvi123, line 118: probably makes more sense to initialize this to false. Otherwise line 122 is redundant

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -13,25 +13,196 @@
...
+                                    WebSocketConstants.DIAGNOSTIC_ON_OPEN_INVALID_PARAMS,
+                                    WebSocketConstants.DIAGNOSTIC_CODE_ON_OPEN_INVALID_PARAMS);
+                            diagnostics.add(diagnostic);
+                            return;
+					    }
+					    
+					    if (!isSpecialType) {
+					        // check that if parameter is not a specialType, it has a @PathParam annotation
+					        IAnnotation[] param_annotations = param.getAnnotations();
+	                        boolean hasPathParamAnnot = Arrays.asList(param_annotations).stream().anyMatch(
```
TheAvi123, line 142: Insignificant issue but formatting seems to be messed up for this code block

TheAvi123 (commented):
A couple of minor things that could be changed but looks good for the most part!

TheAvi123:
Also, I see that we have some constants and such that can be shared between my implementation and yours. Depending on which PR gets merged first, we should take a look to remove duplicates.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/AnnotationUtil.java
dalidia: Good catch, will add that !

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
dalidia: yeah, I missed that part. I will fix that, thanks!

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/websocket/JakartaWebSocketTest.java
```diff
@@ -0,0 +1,66 @@
...
+* Copyright (c) 2021 IBM Corporation and others.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     IBM Corporation - initial API and implementation
```
giancarlopernudisegura, line 11: If you created this file, shouldn't your name be here?

dalidia: Added my name as well

giancarlopernudisegura (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
```diff
@@ -8,14 +8,54 @@
...
 public class WebSocketConstants {
     public static final String DIAGNOSTIC_SOURCE = "jakarta-websocket";
-    public static final DiagnosticSeverity SEVERITY = DiagnosticSeverity.Error;
+
+    public static final DiagnosticSeverity ERROR = DiagnosticSeverity.Error;
+    public static final DiagnosticSeverity WARNING = DiagnosticSeverity.Warning;
+
+    public static final String DIAGNOSTIC_ERR_MSG = "Cannot calculate WebSocket diagnostics";
+
+    public static final String DIAGNOSTIC_PATH_PARAMS_ANNOT_MISSING = "Parameter is missing annotation @PathParams.";
```
kathrynkodama, line 32: Can we rephrase this diagnostic message to be more specific? Indicating that parameters of type String, any Java primitive types or boxed version thereof must be annotated with `@PathParams`

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
```diff
@@ -8,14 +8,54 @@
...
+    public static final String DIAGNOSTIC_CODE_PATH_PARMS_ANNOT = "AddPathParamsAnnotation";
+    
+    public static final String DIAGNOSTIC_ON_OPEN_INVALID_PARAMS = "Invalid parameter type. Parameter must be of type Session, EndpointConfig parameter, or if annotated with @PathParam of type String, any Java primitive type or boxed version thereof.";
```
kathrynkodama, line 35: Let's use full class name in this message, ie. "must be of type jakarta.websocket.Session, ..". The last half of the sentence feels a little clunky. If I am understanding the flow of when each diagnostic shows up, this diagnostic will not show up if on a String or Java primitive type parameter correct? Then maybe we can omit the last piece with `@PathParam` altogether. What do you think?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
dalidia: will do that!

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
dalidia: Yes, you're right. I just pushed a commit with the changes suggested. Thank you!

giancarlopernudisegura: I'm not sure about omitting the @PathParam part as this is how the specification has it but agree on giving out the full name for the classes.

kathrynkodama: Even though that is how it is specified in the spec, if the diagnostic will never show up on a String or Java primitive type parameter (or boxed version thereof) then it is not relevant to include in the message.

dalidia: Yes, that diagnostic message only applies to when the parameter types are checked. If they're not the right ones, that error message is thrown.

kathrynkodama: My understanding of this is that this diagnostic will not show up on any parameter of type String or Java primitive type.
For example, it would not show up on the following (instead the diagnostic would be the one about missing `@PathParam`
```
 public void OnOpen(String stringParam) throws IOException {
```
but would show up on the following:
```
 public void OnOpen(Object objectParam) throws IOException {
```

@dalidia can you please clarify?

kathrynkodama: Ah but in that case you want to indicate to the user all the possible valid types...
How about keeping it as-is, but I wonder if we could display the valid types as a list for readability.
```
    public static final String DIAGNOSTIC_ON_OPEN_INVALID_PARAMS = "Invalid parameter type. Parameter must be of type:
    - jakarta.websocket.Session
    - jakarta.websocket.EndpointConfig
    - annotated with @PathParam and of type String or any Java primitive type or boxed version thereof.";
```
We use a list on a diagnostic message here: https://github.com/eclipse/lsp4jakarta/blob/a195a915af4b7f33d884799927fd786546cd0766/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/beanvalidation/BeanValidationDiagnosticsCollector.java#L194-L198
What do you think?

dalidia: Yes, this diagnostic will not show up for any parameter of type String or Java primitive type, and it will show up in the example provided by Kathryn. The reason why I focused on checking the parameter types first rather than the annotation is because I consider this error more critical from a user's standpoint as this will require some logic change, whereas adding `@PathParam` is a simple fix.

kathrynkodama: Makes sense, thanks @dalidia. In that case I agree with @giancarlopernudisegura that we should keep the `@PathParam` piece in that diagnostic message. What do you think of displaying the diagnostic with a list of supported parameter types.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
```diff
@@ -29,10 +29,10 @@
 
     public static final String DIAGNOSTIC_ERR_MSG = "Cannot calculate WebSocket diagnostics";
 
-    public static final String DIAGNOSTIC_PATH_PARAMS_ANNOT_MISSING = "Parameter is missing annotation @PathParams.";
+    public static final String DIAGNOSTIC_PATH_PARAMS_ANNOT_MISSING = "Parameters of type String, any Java primitive types or boxed version thereof must be annotated with @PathParams.";
```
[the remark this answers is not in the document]
dalidia: I think in this case it should mention `annotated with @PathParams` as this diagnostic message is related to these specific types missing the necessary `@PathParam` annotation. Those parameter types are not considered valid unless they have that annotation. Otherwise, the user will not know why that error is showing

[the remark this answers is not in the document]
kathrynkodama: Yes I agree, sorry full message should be:

    public static final String DIAGNOSTIC_PATH_PARAMS_ANNOT_MISSING = "Parameters of type String, any Java primitive type, or boxed version thereof must be annotated with @PathParams.";

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
dalidia: Sure, I'll do that! Thanks Kathryn and Giancarlo!

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketConstants.java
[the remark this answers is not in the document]
dalidia: will do that, thanks!

kathrynkodama (approved):
Thanks @dalidia, looks good to me.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -8,30 +8,196 @@
...
+    * @param type the type representing the class
+    * @return the conditions for a class to be a WebSocket endpoint 
+    * @throws JavaModelException
+    */
+    private HashMap<String, Boolean> isWSEndpoint(IType type) throws JavaModelException {
+        HashMap<String, Boolean> wsEndpoint = new HashMap<>();
+
+        // check trivial case
+        if (!type.isClass()) {
+            wsEndpoint.put("isAnnotation", false);
```
turkeylurkey, line 179: When you use a string as a key like this it is prone to bugs. Someone could accidentally code `.get("isAnotation")` and introduce a bug which may not be caught. Please put this string in a named constant so that the compiler can check for a misspelt variable name.

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
dalidia: Makes sense, thanks! I just pushed a new change handling it this

turkeylurkey (approved):

PR #259, opened 2022-02-24
Added @Transactional snippet

Jellal-HT:
Resolves #251:
Added @Transactional snippet in the transaction.json file

PR #261, opened 2022-02-24
Auto-complete snippets for UserTransaction and TransactionManager

sullan-kang:


sullan-kang:
Opened #264 about the limitation in implementing this using a JSON snippet, as discussed in the last Thursday's meeting!

However, as I was opening it, it came to my mind that perhaps it would be more appropriate to resolve this using a CodeAction for the following reasons:

- As the user types `utx.begin()`, the statement needs to be surrounded by try/catch along with the generated `utx.commit()`. In other words, try/catch would need to be in the auto-complete snippet, but since "try" comes before "begin", we cannot make "begin" the prefix for the snippet.
- Verifying the type of `utx` is not easy in the snippet method. Perhaps this can be dealt with via CodeAction?
- Using the CodeAction method, we could throw a warning if commit() does not come after begin().

Would it be a feasible idea to implement? @kathrynkodama

sullan-kang:
Thank you, Kathryn. The initialization idea sounds great - I'll get onto it!

As for `.commit` following `.begin`, I cannot find an explicit mention of the requirement. The closest thing I found is the IllegalStateException (https://javadoc.io/doc/jakarta.transaction/jakarta.transaction-api/latest/jakarta/transaction/UserTransaction.html#commit()). I thought perhaps it could be a yellow-underlined quick fix (i.e.., suggestion)? But please let me know what you think!

sullan-kang:
Resolved the [above problem](https://github.com/eclipse/lsp4jakarta/pull/261#issuecomment-1055827566) by including the initialization methods for UserTransaction as @kathrynkodama suggested! 

However, the initialization methods of TransactionManager don't seem to be standardized on [specs](https://jakarta.ee/specifications/transactions/2.0/jakarta-transactions-spec-2.0.html#transactionmanager-interface) so I am wondering if it would be better to leave this PR only relevant to the UserTransactions.

jakarta.ls/src/main/resources/transactions.json
```diff
@@ -1,42 +1,53 @@
...
 			"\ttry {",
-			"\t\ttm.rollback();",
+			"\t\t${1:TransactionName}.rollback();",
 			"\t} catch (Exception e2){",
 			"\t\te2.printStackTrace();",
 			"\t}",
 			"}"
 		],
-		"description": "TransactionManager begin-commit-rollback",
+		"description": "Initializes a UserTransaction Object via JNDI lookup. Provides a template for begin(), commit(), and rollback() of the said UserTransaction.",
```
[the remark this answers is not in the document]
sullan-kang: Fixed!

kathrynkodama:
> However, the initialization methods of TransactionManager don't seem to be standardized on [specs](https://jakarta.ee/specifications/transactions/2.0/jakarta-transactions-spec-2.0.html#transactionmanager-interface) so I am wondering if it would be better to leave this PR only relevant to the UserTransaction.

I think given this it is reasonable to keep the snippets relevant only to the UserTransaction. Can you please modify the title and corresponding issue to clarify that it will not include snippets for TransactionManager?

Jellal-HT:
It looks good to me.

kathrynkodama:
@turkeylurkey please review

jakarta.ls/src/main/resources/transactions.json
```diff
@@ -1,3 +1,53 @@
 {
-
+	"Initialize a UserTransaction object via Injection": {
```
kathrynkodama, line 2: Looks like there may be tabs in this document, to keep consistent with https://github.com/eclipse/lsp4jakarta/pull/259 use spaces instead of tabs

jakarta.ls/src/main/resources/transactions.json
```diff
@@ -1,3 +1,53 @@
...
+		"prefix": "userTransactionInjection",
+		"body": [
+			"@Resource",
+			"UserTransaction ${1:TransactionName};",
```
kathrynkodama, line 6: use camel case for the variables - `transactionName`, `functionName`, etc.

jakarta.ls/src/main/resources/transactions.json
```diff
@@ -1,3 +1,53 @@
...
+		],
+		"description": "Initializes a UserTransaction object via Injection. Provides a template for begin(), commit(), and rollback() of the said UserTransaction object.",
+		"context": {
+			"type": "jakarta.transaction.UserTransaction"
+		}
+	},
+	"Initialize a UserTransaction object via JNDI Lookup": {
+		"prefix": "userTransactionJNDI",
+		"body": [
+			"UserTransaction ${1:TransactionName} = (UserTransaction) context.lookup(\"${2:Address}\");",
```
kathrynkodama, line 32: As discussed we should not reference `context` without a corresponding variable - how is this typically used/initialized?

turkeylurkey: You may be able to find an actual variable of type Context in the Java context where this snippet is used. If there isn't one you could use a default name but it won't compile. We could research how other snippets handle missing dependencies.

sullan-kang:
I opened issue #267 for the indentation problem we saw in today's meeting, as the same behaviour appeared for the snippets in other JSON files.

kathrynkodama:
> I opened issue #267 for the indentation problem we saw in today's meeting, as the same behaviour appeared for the snippets in other JSON files.

Thank you @sullan-kang

jakarta.ls/src/main/resources/transactions.json
sullan-kang: Fixed! Didn't realize, sorry!

jakarta.ls/src/main/resources/transactions.json
sullan-kang: Done!

jakarta.ls/src/main/resources/transactions.json
sullan-kang: I included a Context initialization in the most recent commit. Would it be a possible way to resolve this?

kathrynkodama (approved):
Looks good to me, @sullan-kang just need to resolve the merge conflicts

sullan-kang:
Resolved the merge conflicts!

turkeylurkey (approved):

Issue #242, opened 2022-02-09
JSON Processing: Diagnostic for Json `createPointer` target


jooyoojoo:
#### Description:
Validate the `jsonPointer` String parameter for the `createPointer` method and provide diagnostic information when this JsonPointer target is not valid.

#### Specification:
An empty `jsonPointer` string defines a reference to the target itself. If the `jsonPointer` string is non-empty, it must be a sequence of '/' prefixed tokens.

[JSON Processing Specification](https://jakarta.ee/specifications/jsonp/2.0/apidocs/jakarta.json/jakarta/json/spi/jsonprovider#createPointer(java.lang.String))

#### Example:
Correct Usage 1:
```
JsonPointer jsonPointer = Json.createPointer("/skills/1");
```
Correct Usage 2:
```
JsonPointer jsonPointer = Json.createPointer("");
```
Incorrect Usage:
```
JsonPointer jsonPointer = Json.createPointer("skills/1");
```

#### Type of language feature proposed:
_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

kathrynkodama:
Thanks @jooyoojoo, there may be some implementation overlap/reusability with https://github.com/eclipse/lsp4jakarta/issues/240

PR #263, opened 2022-02-28
Add method to verify URIs have a leading slash

giancarlopernudisegura:
This is needed for #240 and #242.

This PR creates a method in JDTUtils for checking if a string starts with a `/`.


giancarlopernudisegura:
I don't know if I need to add unit tests for this PR but I can do so if needed.

kathrynkodama (approved):
Thanks @giancarlopernudisegura, I don't think we need unit tests for this

Issue #265, opened 2022-03-02
Test projects signature types not resolving correctly


dalidia:
#### Description:
Integration test fails to get the resolved type of a signature when the test Java Project is loaded. 

#### Specification:
As an example of the current test behavior, `JavaModelUtil.getResolvedTypeName()` fails to find the resolved Type (returning null) even though the primaryType and the formatSignature values are correct and are found in the Test Java Project. When doing manual testing, `JavaModelUtil.getResolvedTypeName()` returns the right resolvedTypeName.
 
```java
for (ILocalVariable var : method.getParameters()) {
    String signature = val.getTypeSignature();
    String formatSignature = signature.replace("/", ".");
    IType primaryType = param.getTypeRoot().findPrimaryType();
    String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, primaryType);
}
```

Joseph-Bineesh:
This issue fix is handled in PR : https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/587

PR #275, opened 2022-03-22
add diagnostics for duplicate onMessage methods

### LINKED ISSUE #243: Websocket: Validate number of onMessage method for each WebSocket message formats
#### Description:
Java classes annotated with @ServerEndpoint or @ClientEndpoint may use the @OnMethod annotation. Currently, if a user add more than one method annotated with `@onMessage` handling the same native WebSocket message format (text, binary or pong), it will create a server error, leaving the server unresponsive. However, the user is not notified that this operation is not valid. For this reason, we need to apply a validation of the number of onMessage method for the same native WebSocket message format.

#### Specification:
Valid WebSocket message format are text, binary and pong.

Important to consider that they could be two other @OnMessage methods covering different native WebSocket message formats (text, binary or pong). Example, one @OnMessage handler for binary and another @OnMessage handler for Pong. For more information, the documentation [here](https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html#onmessage)

##### Acceptable
Only a single onMessage method handler for the text format.
```java
@OnMessage
public void validMessage(String message, Session session) {
    System.out.println("Program requested " + message + " using " + session.getId());
    session.getAsyncRemote().sendText("Lucky says hi");
}
```

##### Error
There are 2 onMessage method handler for the text format.
```java
@OnMessage
public void firstOnMessageHandler(String text, Session sessionr) {
   System.out.println("Program requested " + invalidParameter + " using " + session.getId());
   session.getAsyncRemote().sendText("I am pong");
}

@OnMessage
public void secondOnMessageHandler(String text, Session sessionr) {
   System.out.println("Program requested " + invalidParameter + " using " + session.getId());
   session.getAsyncRemote().sendText("I am pong");
}
```
_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

giancarlopernudisegura:
Resolves #243

Show diagnostics if a duplicate method is found for a @onMessage method for the same message format (text, binary, or pong).

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
@@ -132,8 +132,7 @@ public class WebSocketDiagnosticsCollector implements DiagnosticsCollector {
                         // check parameters valid types
                         if (!(isSpecialType || isPrimWrapped || isPrimitive)) {
                             Diagnostic diagnostic = createDiagnostic(param, unit,
-                                    createParamTypeDiagMsg(specialParamTypes, methodAnnotTarget),
-                                    diagnosticCode);
+                                    createParamTypeDiagMsg(specialParamTypes, methodAnnotTarget), diagnosticCode);
                             diagnostics.add(diagnostic);
                             continue;
                         }
@@ -231,24 +230,24 @@ public class WebSocketDiagnosticsCollector implements DiagnosticsCollector {
                                 WebSocketConstants.MESSAGE_FORMAT messageFormat = getMessageFormat(resolvedTypeName);
                                 boolean duplicateFound = false;
                                 switch (messageFormat) {
-                                    case TEXT:
-                                        if (onMessageTextUsed) {
-                                            duplicateFound = true;
-                                        }
-                                        onMessageTextUsed = true;
-                                        break;
-                                    case BINARY:
-                                        if (onMessageBinaryUsed) {
-                                            duplicateFound = true;
-                                        }
-                                        onMessageBinaryUsed = true;
-                                        break;
-                                    case PONG:
-                                        if (onMessagePongUsed) {
-                                            duplicateFound = true;
-                                        }
-                                        onMessagePongUsed = true;
-                                        break;
+                                case TEXT:
+                                    if (onMessageTextUsed) {
+                                        duplicateFound = true;
+                                    }
+                                    onMessageTextUsed = true;
+                                    break;
+                                case BINARY:
+                                    if (onMessageBinaryUsed) {
+                                        duplicateFound = true;
+                                    }
+                                    onMessageBinaryUsed = true;
+                                    break;
+                                case PONG:
+                                    if (onMessagePongUsed) {
+                                        duplicateFound = true;
+                                    }
+                                    onMessagePongUsed = true;
+                                    break;
                                 }
                                 if (duplicateFound) {
                                     Diagnostic diagnostic = createDiagnostic(annotation, unit,
@@ -274,9 +273,7 @@ public class WebSocketDiagnosticsCollector implements DiagnosticsCollector {
             if (annotation.getElementName().equals(WebSocketConstants.SERVER_ENDPOINT_ANNOTATION)) {
                 for (IMemberValuePair annotationMemberValuePair : annotation.getMemberValuePairs()) {
                     if (annotationMemberValuePair.getMemberName().equals(WebSocketConstants.ANNOTATION_VALUE)) {
-                        String path = annotationMemberValuePair
-                                .getValue()
-                                .toString();
+                        String path = annotationMemberValuePair.getValue().toString();
                         Diagnostic diagnostic;
                         if (!JDTUtils.hasLeadingSlash(path)) {
                             diagnostic = createDiagnostic(annotation, unit,
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
@@ -232,7 +232,7 @@ public class WebSocketDiagnosticsCollector implements DiagnosticsCollector {
                             String signature = param.getTypeSignature();
                             String formatSignature = signature.replace("/", ".");
                             String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, type);
-                            if (!resolvedTypeName.equals(WebSocketConstants.SESSION_CLASS)) {
+                            if (resolvedTypeName != null && !resolvedTypeName.equals(WebSocketConstants.SESSION_CLASS)) {
                                 WebSocketConstants.MESSAGE_FORMAT messageFormat = getMessageFormat(resolvedTypeName);
                                 boolean duplicateFound = false;
                                 switch (messageFormat) {
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
@@ -95,6 +95,9 @@ public class WebSocketDiagnosticsCollector implements DiagnosticsCollector {
 
                     // PathParam URI Mismatch Warning Diagnostic
                     uriMismatchWarningCheck(type, diagnostics, unit);
+
+                    // OnMessage validation for WebSocket message formats
+                    onMessageWSMessageFormats(type, diagnostics, unit);
                 }
             }
         } catch (JavaModelException e) {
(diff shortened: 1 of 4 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -204,6 +207,68 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+                                    break;
+                                case PONG:
+                                    if (onMessagePongUsed) {
+                                        duplicateFound = true;
+                                    }
+                                    onMessagePongUsed = true;
+                                    break;
+                                }
+                                if (duplicateFound) {
+                                    Diagnostic diagnostic = createDiagnostic(param, unit,
```
dalidia, line 259: instead of the error showing up in the parameter, would it be better to show up in the method?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -204,6 +207,68 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+            IAnnotation[] allAnnotations = method.getAnnotations();
+            for (IAnnotation annotation : allAnnotations) {
+                if (annotation.getElementName().equals(WebSocketConstants.ON_MESSAGE)) {
+                    ILocalVariable[] allParams = method.getParameters();
+                    for (ILocalVariable param : allParams) {
+                        if (!isParamPath(param)) {
+                            String signature = param.getTypeSignature();
+                            String formatSignature = signature.replace("/", ".");
+                            String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, type);
+                            if (!resolvedTypeName.equals(WebSocketConstants.SESSION_CLASS)) {
```
dalidia, line 235: In the tests, `String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, type)` return null to me. See open issue https://github.com/eclipse/lsp4jakarta/issues/265 . If this issue is still occurring, it would be good to apply a null checker for `resolvedTypeName`

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: Would it be better to have it be the method name or the method annotation?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: Just seeing this comment now but that's essentially what I did to fix the issue after running the debugger.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -199,16 +201,84 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+            IAnnotation[] allAnnotations = method.getAnnotations();
+            for (IAnnotation annotation : allAnnotations) {
+                if (annotation.getElementName().equals(WebSocketConstants.ON_MESSAGE)) {
+                    ILocalVariable[] allParams = method.getParameters();
+                    for (ILocalVariable param : allParams) {
+                        if (!isParamPath(param)) {
+                            String signature = param.getTypeSignature();
+                            String formatSignature = signature.replace("/", ".");
+                            String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, type);
+                            if (resolvedTypeName != null
```
dalidia, line 228: Is there a specific reason why when `resolvedTypeName ` is null, the analysis is ignored? I think we should obtain the parameter type in another way when the `resolvedTypeName` is null?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -228,32 +228,36 @@ private void onMessageWSMessageFormats(IType type, List<Diagnostic> diagnostics,
...
+                                    if (onMessagePongUsed != null) {
+                                        duplicateFound = onMessagePongUsed;
                                     }
-                                    onMessagePongUsed = true;
+                                    onMessagePongUsed = annotation;
                                     break;
                                 }
-                                if (duplicateFound) {
-                                    Diagnostic diagnostic = createDiagnostic(annotation, unit,
+                                if (duplicateFound != null) {
```
dalidia, line 252: It seems that this only handles when there are only 2 onMessage annotations, should we also take care when there are more than 2?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: That was added due to the null pointer exceptions that only happened during the unit tests. See #265

dalidia: I understand that but we should stay obtain the parameter type in another way. I used Signature to get the raw parameter type, maybe you can use something similar to that

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: I could try to generalize it to more than 2 although I don't think it's necessary. Pong message formats only take one class. Text and binary message formats allow for two different classes: String or Reader, and ByteBuffer or InputStream respectively.

This means that if for say we have 3 duplicate @OnMessage text methods, there will be one signature with a string, another with a Reader and the last one with either a String or Reader. This will most likely cause a regular java duplicate method error so they'll get another diagnostic. In the scenario that they changed the signature in another way, that third diagnostic might not appear but as soon as they fix one of the first duplicates, they'd get the message for the second.

dalidia: Makes sense!

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: In what PR was that?

dalidia: See [line 126](https://github.com/eclipse/lsp4jakarta/blob/b49113e3ddd665404441f223d79f198dfdd482bd/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java#L126). PR was [#260](https://github.com/eclipse/lsp4jakarta/pull/260)

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -199,16 +201,97 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+            for (IAnnotation annotation : allAnnotations) {
+                if (annotation.getElementName().equals(WebSocketConstants.ON_MESSAGE)) {
+                    ILocalVariable[] allParams = method.getParameters();
+                    for (ILocalVariable param : allParams) {
+                        if (!isParamPath(param)) {
+                            String signature = param.getTypeSignature();
+                            String formatSignature = signature.replace("/", ".");
+                            String resolvedTypeName = JavaModelUtil.getResolvedTypeName(formatSignature, type);
+                            if (resolvedTypeName == null) {
+                                resolvedTypeName = Signature.getSignatureSimpleName(signature);
```
dalidia, line 229: `Signature.getSignatureSimpleName` does not return the whole class name. Example for String, `JavaModelUtil.getResolvedTypeName` returns `java.lang.String` whereas `Signature.getSignatureSimpleName` returns `String`

dalidia (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: alright thanks for letting me know

dalidia (approved):
Looks good!

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
```diff
@@ -199,16 +201,102 @@ private void uriMismatchWarningCheck(IType type, List<Diagnostic> diagnostics, I
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Create an error diagnostic if a ServerEndpoint annotation does not follow the
```
turkeylurkey, line 289: Please add client endpoint to the comment and also explain the rule a bit.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/websocket/DuplicateOnMessage.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.io.Reader;
+
+import jakarta.websocket.OnMessage;
+import jakarta.websocket.Session;
+import jakarta.websocket.server.PathParam;
+import jakarta.websocket.server.ServerEndpoint;
+
+@ServerEndpoint("/{var}")
+public class DuplicateOnMessage {
```
turkeylurkey, line 11: Could you add a comment to explain that String is the duplicate of Reader?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/websocket/WebSocketDiagnosticsCollector.java
giancarlopernudisegura: I don't know why I added ClientEndpoint to this method as it's not in the spec but this method was part of a previous PR. Regardless, I've removed the client endpoint check and added a rule explanation.

turkeylurkey (approved):

kathrynkodama (approved):

Issue #268, opened 2022-03-04
JSON Binding: JsonbTransient on getters/setters diagnostic


jooyoojoo:
#### Description:
Provide a diagnostic when a getter/setter is annotated with `@JsonbTransient` but the field and this method are annotated with other JSON-B annotations.

#### Specification:
If a getter is annotated with `@JsonbTransient`, exception is thrown if when the field or this getter are annotated with other JSON Binding annotations. Exception is not thrown if JSON Binding annotations are presented on the setter.

If a setter is annotated with `@JsonbTransient`, exception is thrown if when the field or this setter are annotated with other JSON Binding annotations. Exception is not thrown if JSON Binding annotations are presented on the getter.

#### Example:
Acceptable
```java
private String name;

@JsonbTransient
public String getName() {
    return name;
}
```

Error
```java
@JsonbProperty("name")
private String name;

@JsonbTransient
public String getName() {
    return name;
}
```

#### Type of language feature proposed:
_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

adit333:
I'd like to pick this issue

PR #270, opened 2022-03-10
Add diagnostic for getter and setter of a field annotated with JsonbTransient 

jooyoojoo:
Resolves #269 
Create diagnostics for getters and setters of fields annotated with `JsonbTransient`

PR #280, opened 2022-03-30
Diagnostic for Accessor not having mutually exclusive JsonbTransient  annotations

adit333:
Addresses #268

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -19,7 +19,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.function.Function;
 
@@ -43,11 +45,13 @@ import org.eclipse.jdt.core.IAnnotation;
 import org.eclipse.jdt.core.IBuffer;
 import org.eclipse.jdt.core.IClassFile;
 import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.ILocalVariable;
 import org.eclipse.jdt.core.IMember;
 import org.eclipse.jdt.core.IMemberValuePair;
+import org.eclipse.jdt.core.IMethod;
 import org.eclipse.jdt.core.IOpenable;
 import org.eclipse.jdt.core.ISourceRange;
 import org.eclipse.jdt.core.ISourceReference;
@@ -59,6 +63,7 @@ import org.eclipse.jdt.core.SourceRange;
 import org.eclipse.jdt.core.dom.ASTParser;
 import org.eclipse.jdt.core.dom.CompilationUnit;
 import org.eclipse.jdt.core.dom.PackageDeclaration;
+import org.eclipse.jdt.internal.corext.codemanipulation.GetterSetterUtil;
 import org.eclipse.jdt.internal.corext.dom.IASTSharedValues;
 import org.eclipse.lsp4j.Location;
 import org.eclipse.lsp4j.Position;
@@ -624,5 +629,24 @@ public class JDTUtils {
      */
     public static boolean hasLeadingSlash(String uriString) {
         return uriString.startsWith("/");
+    }
+
+    /**
+     * Returns a list of all accessors (getter and setter) of the given field.
+     * 
+     * @param field
+     * @return a list of accessor methods
+     * @throws JavaModelException
+     */
+    public static List<IMethod> getAccessors(IField field) throws JavaModelException {
+        List<IMethod> accessors = new ArrayList<>();
+        IMethod getter = GetterSetterUtil.getGetter(field);
+        IMethod setter = GetterSetterUtil.getSetter(field);
+        for (IMethod accessor : new IMethod[] { getter, setter }) {
+            if (accessor != null) {
+                accessors.add(accessor);
+            }
+        }
+        return accessors;
     }
 }
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -642,9 +642,11 @@ public class JDTUtils {
         List<IMethod> accessors = new ArrayList<>();
         IMethod getter = GetterSetterUtil.getGetter(field);
         IMethod setter = GetterSetterUtil.getSetter(field);
-        for (IMethod accessor : new IMethod[] { getter, setter }) {
-            if (accessor != null) {
-                accessors.add(accessor);
+        if (getter != null && getter.exists() || setter != null && setter.exists()) {
+            for (IMethod accessor : new IMethod[] { getter, setter }) {
+                if (accessor != null) {
+                    accessors.add(accessor);
+                }
             }
         }
         return accessors;
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
@@ -648,9 +648,12 @@ public class JDTUtils {
 
     /**
      * Returns a list of all accessors (getter and setter) of the given field.
+     * Note that for boolean fields the accessor of the form "isField" is retuned
+     * "getField" is not present.
      * 
-     * @param unit, field
-     * @return a list of accessor methods
+     * @param unit      the compilation unit the field belongs to
+     * @param field     the accesors of this field are returned
+     * @return          a list of accessor methods
      * @throws JavaModelException
      */
     public static List<IMethod> getFieldAccessors(ICompilationUnit unit, IField field) throws JavaModelException {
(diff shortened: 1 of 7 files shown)


kathrynkodama:
@adit333 I started debugging some of the failing integration tests and I believe I was able to reproduce them on my machine. I'm running the integration tests with `cd jakarta.jdt && mvn clean install`
I went back to your first commit: https://github.com/eclipse/lsp4jakarta/pull/280/commits/f373b8f013246e6169552f61953bd5381c264304, to me it looks like this is where the tests seemed to start failing. My initial guess is that some of your code changes led to errors in the test project being resolved/uncaught exceptions but may require further investigation.

Re-organizing some of your logic and I believe I resolved the test failure. Please take a look:
https://github.com/kathrynkodama/lsp4jakarta/commit/21a83bca506327b92a16c85d4163c55a13fec3d1

I didn't pull in any commits after your first so may require some more updating, but hopefully it hints at where the error might be. My main changes are in this block (comments denoted with 'K:') 
https://github.com/kathrynkodama/lsp4jakarta/commit/21a83bca506327b92a16c85d4163c55a13fec3d1#diff-86927f1d0ee853909922a8c3d21276e6942f36889fa89ee7f3650a60569d249dR101-R121

The main change was pulling your separate calls to `JDTUtils.getAccessors(field)` under one for loop. 

As as aside, I think we should add a null and exists check to that `getAccessors` method in [JDTUtils](https://github.com/eclipse/lsp4jakarta/blob/2c4c45eb521fb327d6f439f9225e54d1c9f3ac89/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java#L641):
```java
    public static List<IMethod> getAccessors(IField field) throws JavaModelException {
        List<IMethod> accessors = new ArrayList<>();
        IMethod getter = GetterSetterUtil.getGetter(field);
        IMethod setter = GetterSetterUtil.getSetter(field);
        if (getter != null && getter.exists() || setter != null && setter.exists()) {
            for (IMethod accessor : new IMethod[] { getter, setter }) {
                if (accessor != null) {
                    accessors.add(accessor);
                }
            }
        }
        return accessors;
    }
```

Hope this gives you a place to start with and reach out if you have any questions.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
```diff
@@ -634,17 +635,22 @@ public static boolean hasLeadingSlash(String uriString) {
...
-        IMethod setter = GetterSetterUtil.getSetter(field);
-        for (IMethod accessor : new IMethod[] { getter, setter }) {
-            if (accessor != null) {
-                accessors.add(accessor);
+    public static List<IMethod> getFieldAccessors(ICompilationUnit unit, IField field) throws JavaModelException {
+        List<IMethod> accessors = new ArrayList<IMethod>();
+        String fieldName = field.getElementName();
+        fieldName = fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
+        String getterName = "get" + fieldName;
+        String setterName = "set" + fieldName;
```
jooyoojoo, line 647: We should consider the case in which the getter name is "isField" for a boolean field. There was a relevant discussion [here](https://github.com/eclipse/lsp4jakarta/pull/270#discussion_r829438115). We could possibly continue using Util methods.

jooyoojoo:
Overall looks good to me except the comment for `getFieldAccessors`.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
```diff
@@ -634,17 +635,24 @@ public static boolean hasLeadingSlash(String uriString) {
...
-            if (accessor != null) {
-                accessors.add(accessor);
+    public static List<IMethod> getFieldAccessors(ICompilationUnit unit, IField field) throws JavaModelException {
+        List<IMethod> accessors = new ArrayList<IMethod>();
+        String fieldName = field.getElementName();
+        fieldName = fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
+        List<String> accessorNames = new ArrayList<String>();
+        accessorNames.add("get" + fieldName);
+        accessorNames.add("set" + fieldName);
+        accessorNames.add("is" + fieldName);
```
jooyoojoo, line 649: "isField" is the accessor of a field only when the field is a boolean. We can add a check for that.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
adit333: Done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/jsonb/JsonbDiagnosticsCollector.java
```diff
@@ -76,79 +76,76 @@ private void collectJsonbCreatorDiagnostics(ICompilationUnit unit,
...
     }
     
-    private List<String> getJsonbAnnotationNamesForField(IField field) throws JavaModelException {
-        IAnnotation fieldAnnotations[] = field.getAnnotations();
-        List<IAnnotation> fieldAnnotationsList = Arrays.asList(fieldAnnotations);
-        List<String> annotationNames = fieldAnnotationsList.stream().map(
+    private List<String> getJsonbAnnotationNames(IAnnotatable annotable) throws JavaModelException {
+        IAnnotation annotations[] = annotable.getAnnotations();
+        List<IAnnotation> annotationsList = Arrays.asList(annotations);
+        List<String> annotationNames = annotationsList.stream().map(
```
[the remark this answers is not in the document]
adit333: Done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
adit333: Yes, but sometimes, even in the field is boolean the accessor still uses "getField()" instead of "isField()", therefore it might be safer to check for both for edge cases. What do you think? Or should we just check for "isField()" whenever the field is boolean?

adit333:
The `JDTUtils`'s method `getFieldAccessors()` has been modified in order to not use  `GetterSetterUtil` as it is causing build errors in the `jakarta.jdt` component.

More specifically, the build failure only happens when we call `GetterSetterUtil.getSetter()` and does not happen when we call `GetterSetterUtil.getGetter()`

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
jooyoojoo: This is a good point. I did some testing and found that if the field is a boolean, only the method with name `isField` is identified as the getter (i.e. the expected `JsonbException` is thrown). `getField` is not identified as a getter so there is not any exception even if we have a JsonbTransient and another Jsonb annotation on it. 

So I think we should just check for "isField" when the field is boolean.

kathrynkodama:
@adit333 can you please resolve the merge conflicts?

kathrynkodama (approved):

adit333:
Fixed merge conflicts

kathrynkodama (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jsonb/JsonbTransientDiagnostic.java
```diff
@@ -32,18 +30,49 @@ public class JsonbTransientDiagnostic {
...
+    @JsonbNillable 
     private void setId(int id) {
         // A diagnostic is expected on setId because as a setter, it is annotated with other 
         // Jsonb annotations while its corresponding field id is annotated with JsonbTransient
         this.id = id;
     }
-}
+    
+    @JsonbTransient
+    private String getFavoriteDatabse() {
```
turkeylurkey, line 57: This is a negative test case but I think it is going to be affected by the difference in spelling between the field name and the accessor name, "favoriteDatabase" vs "favoriteDatabse" (getFavoriteDatabse).

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jsonb/JsonbTransientDiagnostic.java
```diff
@@ -32,18 +30,49 @@ public class JsonbTransientDiagnostic {
...
+        return favoriteDatabase;
+    }
+    
+    @JsonbTransient
+    private String setFavoriteDatabse(String favoriteDatabase) {
```
turkeylurkey, line 62: This is a negative test case but I think it is going to be affected by the difference in spelling between the field name and the accessor name, "favoriteDatabase" vs "favoriteDatabse" (setFavoriteDatabse).

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
```diff
@@ -647,17 +649,24 @@ public static boolean isValidLevel1URI(String uriString) {
     /**
      * Returns a list of all accessors (getter and setter) of the given field.
      * 
-     * @param field
+     * @param unit, field
```
turkeylurkey, line 652: Normally these Javadoc comments are read by a tool which expects each argument to have its own @param annotation. Even if we don't use this tool, this method is now long enough it makes sense to write a short description of the arguments (one per line).

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jsonb/JsonbTransientDiagnostic.java
adit333: Done. Thank you for catching that! I keep making typos :(

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jsonb/JsonbTransientDiagnostic.java
adit333: Done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTUtils.java
adit333: Done

turkeylurkey (approved):

Issue #300, opened 2022-08-15
Filter JSON diagnostics based on classpath/imports


bensonnw:
#### Description:
- [ ] JsonbDiagnosticsCollector
- [ ] JsonpDiagnosticCollector

#### Type of language feature proposed:
_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

PR #316, opened 2022-09-02
Validate JSON annotations

bensonnw:
#300


bensonnw:

PR #322, opened 2022-09-20
Convert Windows style line endings to Unix style.

turkeylurkey:
Converted line endings on all lines even though github only shows them on certain lines.


turkeylurkey:
@kathrynkodama  @bensonnw could you take a look and review?

kathrynkodama (approved):

Issue #295, opened 2022-07-26
Buildable jar for the lsp4jakarta.jdt and lsp4jakarta.ls components for clients to consume


kathrynkodama:
Produce jars of the `lsp4jakarta.jdt` (JDT LS extension) and `lsp4jakarta.ls` components that IDE clients can consume in an automated build script

PR #341, opened 2022-10-18
DelegateCommandHandler code for clients to consume LSP4Jakarta

kathrynkodama:
Related to #295 

Necessary fixes for IDE clients to consume LSP4Jakarta. 
The JakartaDelegateCommandHandlerForJava is required for IDE clients who do not implement the JakartaLanguageClientAPI

Co-authored by @dshimo


jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaTextDocumentService.java
```diff
@@ -156,15 +170,24 @@ private void triggerValidationFor(List<String> uris) {
...
-                });
+
+        // Async thread to query the JDT LS ext for Jakarta EE diagnostics
+        CompletableFuture<List<PublishDiagnosticsParams>> getJakartaDiagnostics = CompletableFuture.supplyAsync(() -> {
+            try {
+                // Pass the JakartaDiagnosticsParams to IDE client, to be forwarded to the JDT
+                // LS ext
+                return jakartaLanguageServer.getLanguageClient().getJavaDiagnostics(javaParams).get();
+            } catch (Exception e) {
+                LOGGER.severe("Return request from client did not succeed: " + e.getMessage());
```
dshimo, line 181: A minor comment that it may be beneficial to distinguish the two new log messages by adding request type
I think i.e. `Return diagnostic request ...` or `Return completion request ...`

dshimo (approved):
Looks good
Just one minor comment on logging

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaTextDocumentService.java
kathrynkodama: Good idea, added in the request type

bensonnw (approved):

Issue #43, opened 2020-11-05
Implement dynamic templating in completionSnippets


AnkushSharma2698:
Certain key variables should resolve automatically in snippets.
Example:
```
${package_name}
```
Should resolve automatically based on the context to
```
package com.example.system;
```

I just did some testing, and the proposed method doesnt actually work in eclipse rn. I tried messing around with templates in eclipse and that does fill in snippets with variables as we want but for whatever reason that is not extending to our current completion framework. So for now I would say we should keep snippets like this. In order to accomodate the change in this request I think i would have to restructure completionProposalComputer a bit and I am not sure about what is needed for that yet

bensonnw:
What `dynamic templating` are we looking for here? Current `completionSnippets` code is in LS side, there is no JDT information available, we need APIs to communicate with client. 
There are many variables defined in JSON files, `${1:packagename}`, `${2:path}`, `${3:className}`, `${4:methodname}` etc. we only need to resolve package and class names?

PR #352, opened 2022-10-26
Resolve snippet package and class name variables

bensonnw:
#43 

Signed-off-by: bensonnw <benson.ning@gmail.com>

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
@@ -165,6 +165,17 @@ public class JDTServicesManager {
                 classpath.add(null);
             });
         }
+
+        // FOR NOW, append package name and class name to the list in order for LS to
+        // resolve ${packagename} and ${classname} variables
+        String className = unit.getElementName();
+        if (className.endsWith(".java") == true) {
+            className = className.substring(0, className.length() - 5);
+        }
+        String packageName = unit.getParent() != null ? unit.getParent().getElementName() : "";
+        classpath.add(packageName);
+        classpath.add(className);
+
         return classpath;
     }
 
--- a/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
+++ b/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
@@ -407,6 +407,14 @@ public class SnippetRegistry {
         return expr.toString();
     }
 
+    /**
+     * Get all matched variables from given string line.
+     * 
+     * @param line    - given string to search
+     * @param start   - position/index to start the search from
+     * @param vars    - searching variables
+     * @param matched - found variables
+     */
     private static void getMatchedVariables(String line, int start, String[] vars,
             Map<String, Set<String>> matched) {
         int idxS = line.indexOf("${", start);
--- a/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
+++ b/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
@@ -27,7 +27,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import org.eclipse.lsp4j.CompletionItem;
@@ -56,8 +55,6 @@ public class SnippetRegistry {
     private static final String PACKAGE_NAME = "packagename";
     private static final String CLASS_NAME = "classname";
     private static final String[] RESOLVE_VARIABLES = { PACKAGE_NAME, CLASS_NAME };
-
-    private static final Logger LOGGER = Logger.getLogger(SnippetRegistry.class.getName());
 
     List<Snippet> snippets; // Hold all snippets in this list
 
--- a/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
+++ b/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
@@ -20,7 +20,14 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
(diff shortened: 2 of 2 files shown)


bensonnw:

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
```diff
@@ -165,6 +165,17 @@ public List<String> getExistingContextsFromClassPath(String uri, List<String> sn
                 classpath.add(null);
             });
         }
+
```
turkeylurkey, line 168: I believe if there is an exception caught at line 158 then the list will be one element short. This would affect the length calculation used later. I think you need to add a null to the output list if you catch an exception.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
```diff
@@ -165,6 +165,17 @@ public List<String> getExistingContextsFromClassPath(String uri, List<String> sn
                 classpath.add(null);
             });
         }
+
+        // FOR NOW, append package name and class name to the list in order for LS to
+        // resolve ${packagename} and ${classname} variables
+        String className = unit.getElementName();
```
kathrynkodama, line 171: Do we need a null check here?

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
```diff
@@ -165,19 +176,27 @@ public List<Snippet> getSnippets() {
      */
     public List<CompletionItem> getCompletionItem(final Range replaceRange, final String lineDelimiter,
             boolean canSupportMarkdown, List<String> context) {
+        List<Snippet> snippets = getSnippets();
+        Map<String, String> values = new HashMap<String, String>();
+        int size = context.size();
+        if (size == snippets.size() + 2) { // the last 2 strings are package name and class name
+            values.put(PACKAGE_NAME, context.get(size - 2));
+            values.put(CLASS_NAME, context.get(size - 1));
```
kathrynkodama, line 184: Should we remove these from the snippet list before streaming?

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
```diff
@@ -367,4 +406,24 @@ protected static String findExprBeforeAt(String text, int offset) {
         }
         return expr.toString();
     }
+
+    private static void getMatchedVariables(String line, int start, String[] vars,
```
kathrynkodama, line 410: Suggestion for comment here, it wasn't immediately clear to me what this method does.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
bensonnw: It's the code from before, not part of changes from here, but I agree to add null to keep the list safe when exception happens (even I doubt the chance, the ctx is pre-defined from JSON files, should be always valid)

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
bensonnw: getElementName from a valid `ICompilationUnit` should never be null, it's JavaElement, it should always have a name, it's very common API and used everywhere

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
bensonnw: Before streaming snippets? are you talking about removing package and class name from context list?

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
bensonnw: hmm..., usually no comments for private method, the API is finding the matched variables from snippet text, bad method name or you want comments on implementation?

turkeylurkey: I was also confused about the int start variable and the subsequent recall so a few sentences would be nice.

bensonnw: If you can understand `replace` API call, It's a pretty similar idea. `start` is the start position of snippet text that the search from.

bensonnw: Anyway, as you guys preferred, I added a comment to the method

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/SnippetRegistry.java
kathrynkodama: Yes from the context list

bensonnw: why? they won't impact to the following usage of the list, not necessary to do extra removal.

kathrynkodama: A suggestion to make the implementation cleaner, but if you feel it is not needed that is okay

kathrynkodama (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
turkeylurkey: Benson were you going to add the `classpath.add(null);` in the catch?

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JDTServicesManager.java
bensonnw: no, it's better to keep that one out of this PR for better history tracing. You can open up a separate issue for your found

Issue #347, opened 2022-10-25
LS throws NPE when closing the last Java file in Eclipse


bensonnw:
#### Description:
Open a Java file in Eclipse, (make sure it's the only one in editor), close it, there is NullPointerException from LS

```
Oct. 25, 2022 10:36:54 A.M. org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer fireError
SEVERE: Cannot invoke "java.util.concurrent.CompletableFuture.thenAccept(java.util.function.Consumer)" because "future" is null
java.lang.NullPointerException: Cannot invoke "java.util.concurrent.CompletableFuture.thenAccept(java.util.function.Consumer)" because "future" is null
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:279)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

PR #361, opened 2022-11-01
Fix NPE when shutting down LS

bensonnw:
#347 
Signed-off-by: bensonnw <benson.ning@gmail.com>

### DIFF
--- a/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
+++ b/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
@@ -21,6 +21,7 @@ import org.eclipse.lsp4j.InitializeParams;
 import org.eclipse.lsp4j.InitializeResult;
 import org.eclipse.lsp4j.ServerCapabilities;
 import org.eclipse.lsp4j.TextDocumentSyncKind;
+import org.eclipse.lsp4j.jsonrpc.CompletableFutures;
 import org.eclipse.lsp4j.services.LanguageClient;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.lsp4j.services.TextDocumentService;
@@ -70,8 +71,10 @@ public class JakartaLanguageServer implements LanguageServer, ProcessLanguageSer
 
     @Override
     public CompletableFuture<Object> shutdown() {
-        // TODO Auto-generated method stub
-        return null;
+        // when shutting down LS, TextDocumentService.didClose() may not be called
+        // properly, need to clear existing diagnostics
+        ((JakartaTextDocumentService) textDocumentService).cleanDiagnostics();
+        return CompletableFutures.computeAsync(cc -> new Object());
     }
 
     @Override
--- a/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaTextDocumentService.java
+++ b/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaTextDocumentService.java
@@ -153,7 +153,6 @@ public class JakartaTextDocumentService implements TextDocumentService {
                         return e;
                     }).collect(Collectors.toList());
                 });
-
     }
 
     // diagnostic request
@@ -172,7 +171,7 @@ public class JakartaTextDocumentService implements TextDocumentService {
         javaParams.setDocumentFormat(DocumentFormat.Markdown);
 
         // Async thread to query the JDT LS ext for Jakarta EE diagnostics
-        CompletableFuture<List<PublishDiagnosticsParams>> getJakartaDiagnostics = CompletableFuture.supplyAsync(() -> {
+        CompletableFuture.supplyAsync(() -> {
             try {
                 // Pass the JakartaDiagnosticsParams to IDE client, to be forwarded to the JDT
                 // LS ext
@@ -188,6 +187,14 @@ public class JakartaTextDocumentService implements TextDocumentService {
                 jakartaLanguageServer.getLanguageClient().publishDiagnostics(diagnostic);
             }
             return jakartaDiagnostics;
+        });
+    }
+
+    protected void cleanDiagnostics() {
+        // clear existing diagnostics
+        documents.all().forEach(doc -> {
+            jakartaLanguageServer.getLanguageClient()
+                    .publishDiagnostics(new PublishDiagnosticsParams(doc.getUri(), new ArrayList<Diagnostic>()));
         });
     }
 }


bensonnw:

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
```diff
@@ -70,8 +71,10 @@ public synchronized void updateSettings(Object initializationOptionsSettings) {
 
     @Override
     public CompletableFuture<Object> shutdown() {
-        // TODO Auto-generated method stub
-        return null;
+        // when shutting down LS, TextDocumentService.didClose() may not be called
+        // properly, need to clear existing diagnostics
+        ((JakartaTextDocumentService) textDocumentService).cleanDiagnostics();
```
turkeylurkey, line 76: Why do we need to publish diagnostics when we are shutting down?

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
```diff
@@ -70,8 +71,10 @@ public synchronized void updateSettings(Object initializationOptionsSettings) {
...
+        return CompletableFutures.computeAsync(cc -> new Object());
```
turkeylurkey, line 77: Usually we use `()` when there are no arguments. A variable name may imply something is missing.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
bensonnw: it's not publishing diagnostics, it's cleaning up existing diagnostics.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
bensonnw: not sure what do you mean

turkeylurkey: `cc -> new Object()` defines an anonymous class with a method that takes the argument `cc` but that argument is not used in the body of the method `new Object()`. This can be confusing when trying to maintain the code especially when the method is longer, non-trivial. Instead, the usual way to code a lambda that does not take any arguments is `() -> new Object()'. This way people do not look for where the argument is used and they don't need to wonder if the part of the method that uses an argument like `cc` has been accidentally erased.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
turkeylurkey: I see that it is called `cleanDiagnostics` but the code it runs is `jakartaLanguageServer.getLanguageClient().publishDiagnostics(new PublishDiagnosticsParams(doc.getUri(), new ArrayList<Diagnostic>())` basically just `publishDiagnostics`. I wasn't able to find the source for `publishDiagnostics`.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
bensonnw: I don't know how much confusing this one line code can cause here, if you are talking about changing something like `() -> new Object()`, unfortunately, in Java, you can NOT ignore pre-defined argument:
image

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
bensonnw: no, `cleanDiagnostics` is LS private method.

bensonnw: not sure what do you mean by "I wasn't able to find the source for publishDiagnostic", that's the API pushing the diagnostic messages to edtior UI, there must be one in IntelliJ, otherwise, you won't see those messages in IntelliJ

bensonnw: by the way, can we do this kind of conversation at some other place (e.g. Slack)? instead of holding this PR to be merged.

bensonnw: One more thing to be clarify here, `cleanDiagnostics` is the `protected` API specific to `JakartaTextDocumentService`, not defined for generic `TextDocumentService` interface.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
turkeylurkey: Ok, sorry, I didn't see that message when I tried it.

turkeylurkey: Heh, just realized CompletableFutures is not CompletableFuture. Tricky. Rookie mistake.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
turkeylurkey: My point is that if the `JakartaTextDocumentService` needs this ability most likely the MicroProfile language server needs this ability also since they do mostly the same thing. That is why I'm wondering if this ability is needed in the more general interface. Your use case seemed like it could happen with a MicroProfile document as well as this type.

turkeylurkey (approved):

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaLanguageServer.java
bensonnw: I don't know whether MP has this issue or not, but I do see the issue in Jakarta LS, if you concern the differences between MP and Jakarta, probably open up an issue about that and do investigation after MVP?

bensonnw: Are you concerning extra "cleanup" steps being called? if you look at the code closely, the `cleanDiagnostics` is just doing a "back-up" step, if there is no message left over, the API will do nothing.

Issue #350, opened 2022-10-26
Document what is supported in which release of LSP4Jakarta


yeekangc:
Document or at least summarise what capabilities (coding assistance) are supported in which release of LSP4Jakarta.

e.g., translate what is captured in https://github.com/eclipse/lsp4jakarta/issues/16 to an end-user (developer) consumable information mapping to the corresponding release(s) of the language server.

TrevCraw:
@yeekangc Is this necessary to document before d-cut? If so, work on this will have an impact on d-cut development work.

yeekangc:
I believe we should document what is relevant as best we can for each release we put out there.

Yes, understand that we are working through various development items.  As we get to more established development schedule and release cadence for the project, documentation or description for what is available in each release (0.x or not) should be part of the release cycle.

kathrynkodama:
Documentation for snippets was added with #365

kathrynkodama:
Documentation for diagnostics & quick-fixes are still outstanding, and should be added to the README: https://github.com/eclipse/lsp4jakarta#jakarta-ee-api-language-features

kathrynkodama:
Closed with https://github.com/eclipse/lsp4jakarta/pull/417

PR #388, opened 2022-12-05
Update readme.md

kathrynkodama:


kathrynkodama:
@TrevCraw please review

README.md
```diff
@@ -24,9 +25,14 @@ This project contains:
 - [Contributing](#Contributing)
 - [Feedback](#Feedback)
 --- 
+## Client IDEs
+
+Eclipse LSP4Jakarta is consumed by the following IDEs:
+-  Eclipse IDE with [Liberty Tools for Eclipse](https://github.com/OpenLiberty/liberty-tools-eclipse#welcome-to-the-liberty-tools-for-eclipse-project)
```
TrevCraw, line 31: should we mention that it is the early release?

README.md
```diff
@@ -24,9 +25,14 @@ This project contains:
...
+
 ## Jakarta EE API Language Features
 
-The following is a full list of Eclipse LSP4Jakarta features offered for each [Jakarta EE API](https://jakarta.ee/specifications/). Language features will only be delivered for a given Jakarta EE API if that API is found on the project's classpath. Eclipse LSP4Jakarta is currently targeting Jakarta EE 9/9.1 (jakarta.x namespace). 
+The following is a list of Eclipse LSP4Jakarta snippets offered for each [Jakarta EE API](https://jakarta.ee/specifications/). For a full list of langauge features (snippets, diagnostics and quick-fixes) currently offered by Eclipse LSP4Jakarta see [issue #16](https://github.com/eclipse/lsp4jakarta/issues/16). Language features will only be delivered for a given Jakarta EE API if that API is found on the project's classpath. Eclipse LSP4Jakarta is currently targeting Jakarta EE 9/9.1 (jakarta.x namespace).
```
TrevCraw, line 35: are the current offered features just the ones that are checked off in that issue? If so, maybe mention that. (I also see some unchecked boxes next to closed issues - should those be updated as well?)

README.md
kathrynkodama: I don't think we need to call out here that Liberty Tools for Eclipse is an early release, it is at the top of their readme

README.md
kathrynkodama: Yes that's all the documentation we have right now until we address #350. I've checked off all the closed issues.

Issue #423, opened 2023-02-21
NullPointerException at org.eclipse.lsp4jakarta.JakartaTextDocumentService.didChange


TrevCraw:
This NPE has been occurring repeatedly in the VS Code IDE. It is not clear if the root cause (how "document" becomes null) is on the client side or the LSP4Jakarta side, however, the NPE needs to be handled on the LSP4Jakarta side. The stack trace shows that the NPE is occurring on this line: https://github.com/eclipse/lsp4jakarta/blob/8e4bd24af7a2dab59a5937ba72daab390af91aa5/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/JakartaTextDocumentService.java#L75

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify(GenericEndpoint.java:152)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification(RemoteEndpoint.java:220)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:187)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
[...]

PR #448, opened 2023-04-26
Check document for null before use.

turkeylurkey:
Investigated related methods but none of them return nulls which are used.


mattbsox (approved):
Looks good to me!

bensonnw (approved):

turkeylurkey:
Fixes #423

PR #452, opened 2023-05-24
LS side for context-aware snippets

evie-lau:


jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/api/JakartaLanguageClientAPI.java
```diff
@@ -49,6 +50,16 @@ default CompletableFuture<List<PublishDiagnosticsParams>> getJavaDiagnostics(Jak
     default CompletableFuture<List<String>> getContextBasedFilter(JakartaClasspathParams classpathParams) {
         return CompletableFuture.completedFuture(null);
     }
+
```
turkeylurkey, line 53: I don't think microprofile/java/javaCursorContext and jakarta/java/completion messages are required so we can remove these two entries.

turkeylurkey (changes requested):
Just one bit I don't think we need, thank you!

turkeylurkey (approved):

PR #458, opened 2023-05-31
Remove default method impls on JakartaLanguageClientAPI intf

### LINKED ISSUE #457: Would it be better to not use default methods in JakartaLanguageClientAPI to force clients to pick up changes?
So if we look at the recent changes: 

https://github.com/eclipse/lsp4jakarta/pull/452
https://github.com/eclipse/lsp4jakarta/pull/453

from the perspective of Liberty Tools Eclipse, the changes completely broke the existing ability to use snippets as completion proposals.

The fix was as simple as adding the new method to our JakartaLanguageClientAPI impl in Liberty Tools Eclipse:
https://github.com/OpenLiberty/liberty-tools-eclipse/pull/387/files

If it weren't for the use of default methods in: https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/api/JakartaLanguageClientAPI.java
this would've been easily detected as a compile-time error... instead we had to go through a debug process.

Is it really helpful to try to support older "clients" with newer levels of LSP4Jakarta in a case like this by adding the default method?    Or is it better to assume a client depends on a specific version of LSP4Jakarta, and to communicate that the new client needs to implement this interface?

scottkurz:
Fixes #457 .  @turkeylurkey

### DIFF
--- a/jakarta.eclipse/org.eclipse.lsp4jakarta.lsp4e.core/src/org/eclipse/lsp4jakarta/lsp4e/JakartaLanguageClient.java
+++ b/jakarta.eclipse/org.eclipse.lsp4jakarta.lsp4e.core/src/org/eclipse/lsp4jakarta/lsp4e/JakartaLanguageClient.java
@@ -85,6 +85,7 @@ public class JakartaLanguageClient extends LanguageClientImpl implements Jakarta
         });
     }
 
+    @Override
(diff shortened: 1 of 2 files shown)


scottkurz:
Looks like this is fixed.  Thanks

Issue #305, opened 2022-08-16
Improving automated tests for diagnostics and quick-fixes


kathrynkodama:
Identifying gaps in the current automated tests given #54 

The update uses fully qualified names so need to test if the package name is incorrect.

TrevCraw:
The main goal of this issue is to create tests to verify we are using the fully qualified names to provide diagnostics.

Create new Java files in the [jakarta-sample project](https://github.com/eclipse/lsp4jakarta/tree/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample) which contain fake package imports with Jakarta API package endings. For example, change the `jakarta.ws.rs.DELETE` package in [MultipleEntityParamsResourceMethod.java](https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/jax_rs/MultipleEntityParamsResourceMethod.java) to `fake.ws.rs.DELETE`. 

The test should then verify that we do NOT provide a diagnostic on the `resourceMethodWithTwoEntityParams` method. Check the `org.eclipse.lsp4jakarta.jdt.core.JakartaForJavaAssert.*;` package to see if there are any methods that can help with confirming there is no diagnostic. If not, we may need to come up with a new method or strategy to check for no diagnostic.

TrevCraw:
1)We should create new tests specifically just for checking for no diagnostics.
2) The test should attempt to grab a list of all the diagnostics in the "test project" file and confirm that the list of diagnostics is empty. There should not be any checks for specific diagnostics.
3) Create a new project with all the diagnostics that were previously using the "short name" check. This might apply for all diagnostics. It may be necessary to copy the entire existing test project and modify the package imports. It is OK if the new project does not compile.
4) Loop the new test checking for no diagnostics to run against every file in the new test project.

PR #483, opened 2023-06-28
#305 Improving automated tests for diagnostics and quick-fixes 

aparnamichael:
Added test case for verifying whether we are using the fully qualified name to provide diagnostics.


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JakartaForJavaAssert.java
```diff
@@ -156,22 +156,32 @@ public static Range r(int startLine, int startCharacter, int endLine, int endCha
     public static Position p(int line, int character) {
         return new Position(line, character);
     }
+  
+    public static void assertJavaDiagnosticsPresent(JakartaDiagnosticsParams params, JDTUtils utils, Diagnostic... expected)
```
turkeylurkey, line 160: Since this method is going to assert a diagnostic is not present it should be renamed assertJavaDiagnosticsNotPresent().

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JakartaForJavaAssert.java
```diff
@@ -156,22 +156,32 @@ public static Range r(int startLine, int startCharacter, int endLine, int endCha
...
 
-    public static void assertDiagnostics(List<Diagnostic> actual, Diagnostic... expected) {
-        assertDiagnostics(actual, Arrays.asList(expected), false);
+    public static void assertDiagnostics( boolean checkForDiagnostic, List<Diagnostic> actual, Diagnostic... expected) {
+        assertDiagnostics(actual, Arrays.asList(expected), false, checkForDiagnostic);
     }
 
-    public static void assertDiagnostics(List<Diagnostic> actual, List<Diagnostic> expected, boolean filter) {
-        /**
+    public static void assertDiagnostics(List<Diagnostic> actual, List<Diagnostic> expected, boolean filter, boolean isDiagnosticPresent) {
```
turkeylurkey, line 182: It is confusing when the argument is called checkForDiagnostic in one method and is called something different here. I suggest you rename this argument to `checkForDiagnostic`.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/core/JakartaForJavaAssert.java
```diff
@@ -212,9 +222,13 @@ public static void assertDiagnostics(List<Diagnostic> actual, List<Diagnostic> e
                 return simpler;
             }).collect(Collectors.toList());
         }
-        Assert.assertEquals("Unexpected diagnostics:\n" + actual, expected, received);
+        if (isDiagnosticPresent) {
```
turkeylurkey, line 225: It is confusing when `checkForDiagnostic` actually means not to check for it. Please reverse the sense of this check so that the else part is after the if and the if part is moved to the else part. Also you will need to change the call sites (l. 164 and 173) so that it still works.

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/nodiagnostics/NoDiagnosticsTest.java
```diff
@@ -0,0 +1,118 @@
...
+		assertJavaDiagnostics(diagnosticsParams, utils);
+
+	}
+
+	// list of java files in demo-servlet-no-diagnostics.
+	@Parameters
+	public static String[] projectFileProvider() throws Exception {
+
+		String basePath = "src/main/java/io/openliberty/sample/jakarta/";
+		String files[] = { basePath + "annotations/GeneratedAnnotation.java",
```
TrevCraw, line 63: I think we should see if there is a way to dynamically retrieve the Java files from the "no-diagnostics" project. It will avoid having to modify this file every time new diagnostic test project files are added.

turkeylurkey (approved):
Other than Trevor's suggestion for improvement it looks good.

turkeylurkey:
The failure in the tests is: 
```
Caused by: java.io.FileNotFoundException: /home/runner/work/lsp4jakarta/lsp4jakarta/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/target/workingProjects/jakarta-noDiagnostic-sample/.project (No such file or directory)```
Possibly copying the .project from jakarta-sample could work but need to examine if the directory name is in the file.

File: jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/nodiagnostics/NoDiagnosticsTest.java [hunk 5]
```diff
@@ -0,0 +1,88 @@
...
+
+		String packagePath = "src/main/java/io/openliberty/sample/jakarta/";
+		String basePath = System.getProperty("user.dir")
+				+ "/projects/demo-servlet-no-diagnostics/" + packagePath;
+
+		List<String> results = new ArrayList<String>();
+		File[] directories = new File(basePath).listFiles();
+		// If this pathname does not denote a directory, then listFiles() returns null.
+
+		for (File directory : directories) {
```
TrevCraw, line 73: This strategy does not account for changes to the structure of the project. What if the directories eventually contain subdirectories? Or a Java file is added to the base directory? And we also need to filter for Java files, not return just any file.

I would recommend using a library that recursively retrieves all files, including files in subdirectories, and filters based on a `java` extension. The [listFiles()](https://commons.apache.org/proper/commons-io/javadocs/api-2.7/org/apache/commons/io/FileUtils.html#listFiles-java.io.File-java.lang.String:A-boolean-) method from Apache Commons IO would work great. You can specify `java` as the file extension to match on and set the `recursive` boolean to `true`.

aparnamichael: @TrevCraw - I used Apache Commons IO's listFiles() method. Tested with subdirectories also it is working fine.

TrevCraw (approved):

TrevCraw:
@aparnamichael did you try running a negative test to confirm that the test fails if Jakarta diagnostics are found in one of the test project files?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/nodiagnostics/NoDiagnosticsTest.java
```diff
@@ -0,0 +1,81 @@
...
+	public static List<String> projectFileProvider() throws Exception {
+
+		String packagePath = "./src/main/java/io/openliberty/sample/jakarta/";
+		String basePath = System.getProperty("user.dir")
+				+ "/projects/demo-servlet-no-diagnostics/";
+		File dir = new File(basePath + packagePath);
+		String[] extensions = new String[] { "java" };
+		List<String> results = new ArrayList<String>();
+
+		List<File> files = (List<File>) FileUtils.listFiles(dir, extensions, true);
```
turkeylurkey, line 73:
```suggestion
		Collection<File> files = FileUtils.listFiles(dir, extensions, true);
```
It is often unwise to cast a type. In this case someone may change the version number of the dependency and the new `FileUtils.listFiles()` returns a different Collection and this code would fail. The code works just as well without casting the type.

turkeylurkey (changes requested):

aparnamichael:
> @aparnamichael did you try running a negative test to confirm that the test fails if Jakarta diagnostics are found in one of the test project files?

Yes tested the negative scenario. User will get an error message in the same format like diagnostic test cases failure.

Failures:
 NoDiagnosticsTest.checkForNoDiagnostics:58 Unexpected diagnostics: 
 expected:<[]> but was:<[Diagnostic [
 range = Range [
  start = Position [
   line = 13
   character = 4
  ] .... ]

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/nodiagnostics/NoDiagnosticsTest.java
aparnamichael: @TrevCraw - Removed the type casting to List<Files>.

turkeylurkey (approved):

TrevCraw (approved):

turkeylurkey:
Fixes #305

PR #492, opened 2023-08-14
Properly handle a non-static field with the @Dependent annotation.

### LINKED ISSUE #444: Quick fix for invalid Managed Bean scope `@DEPENDENT` does not resolve diagnostic
The public field in this sample taken from `jakarta-sample/src/main/java/io/openliberty/sample/jakarta/cdi/ManagedBean.java` provides a quick fix to change `@RequestScoped` into `@DEPENDENT` but this does not resolve the diagnostic on the field itself. 
image

We either need another quick fix to add an annotation to the field or fix the diagnostic not to flag the field when the class has annotation `@DEPENDENT`.

mezarin:


mezarin:
Closing this PR because the code base has changed. This PR is being replaced by #506

Issue #134, opened 2021-04-01
Add quick-fix for Jakarta CDI Produces, Inject, Disposes, Observes/ObservesAsync diagnostics


Jnnnx7:
- [x] **A bean constructor or an initializer method annotated with `@Inject` cannot have a parameter annotated `@Disposes`, `@Observes`, or `@ObservesAsync`**
See the issue for corresponding diagnostic here: https://github.com/eclipse/lsp4jakarta/issues/104

Add two quick-fixes for this diagnostic:
1. Remove the `@Inject` annotation
2. Remove the conflicted annotations (`@Disposes`, `@Observes`, and/or `@ObservesAsync`)<br/><br/>

- [x] **A producer method cannot have a parameter annotated `@Disposes`, `@Observes`, or `@ObservesAsync`**
See the issue for corresponding diagnostic here: https://github.com/eclipse/lsp4jakarta/issues/105

Add two quick-fixes for this diagnostic:
1. Remove the `@Produces` annotation
2. Remove the conflicted annotations (`@Disposes`, `@Observes`, and/or `@ObservesAsync`)<br/><br/>

- [x] **A method cannot have more than one parameter annotated `@Disposes`**
See the issue for corresponding diagnostic here: https://github.com/eclipse/lsp4jakarta/issues/106

Add a quick-fix to remove the extra `@Disposes` annotations<br/><br/>

- [x] **A disposer method cannot have a parameter annotated @Observes or a parameter annotated @ObservesAsync**
See the issue for corresponding diagnostic here: https://github.com/eclipse/lsp4jakarta/issues/106

Add two quick-fixes for this diagnostic:
1. Remove the `@Disposes` annotation
2. Remove the conflicted annotations (`@Observes`, and/or `@ObservesAsync`)

ajm01:
fixed in #511

PR #511, opened 2024-01-16
fix issue 134 - add quick fixes for CDI Produces and Disposes diagnos…

ajm01:
issue 134 - add missing code action quick fixes for existing @Produces, @Disposes and @Inject related diagnostics.

https://github.com/eclipse/lsp4jakarta/issues/134

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/plugin.xml
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/plugin.xml
@@ -297,15 +297,21 @@
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.InsertDefaultPublicConstructorToMBeanQuickFix" />
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidInjectAnnotatedMethodParamAnnotation"
-                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectParamAnnotationQuickFix" />    
+                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInitializerParamAnnotationQuickFix" />    
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidInjectAnnotatedMethodParamAnnotation"
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInjectAnnotationQuickFix" />   
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidProducerMethodParamAnnotation"
-                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectParamAnnotationQuickFix" />     
+                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidProducerParamAnnotationQuickFix" />     
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidProducerMethodParamAnnotation"
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveProduceAnnotationQuickFix" />
+      <codeAction kind="quickfix"
+                  targetDiagnostic="jakarta-cdi#InvalidDisposesAnnotationOnMultipleMethodParams"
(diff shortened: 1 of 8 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/Constants.java
```diff
@@ -48,9 +48,9 @@ public class Constants {
 
     public static final String DIAGNOSTIC_CODE_REDUNDANT_DISPOSES = "RemoveExtraDisposes";
 
-    public static final Set<String> INVALID_INJECT_PARAMS = new HashSet<String>(Arrays.asList(DISPOSES, OBSERVES, OBSERVES_ASYNC));
-    public static final String[] INVALID_INJECT_PARAMS_FQ = { DISPOSES_FQ_NAME, OBSERVES_FQ_NAME,
-                                                              OBSERVES_ASYNC_FQ_NAME };
+    public static final Set<String> INVALID_CDI_PARAMS = new HashSet<String>(Arrays.asList(DISPOSES, OBSERVES, OBSERVES_ASYNC));
```
mezarin, line 51: I don't think reusing the variable is the way to go because it becomes confusing and hard to read. I think there should be the list of invalid parameters for initializer methods and producer methods, which are the things that are being handled here. For example: INVALID_INITIALIZER_METHOD_PARAMS, INVALID_PRODUCER_METHOD_PARAMS

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/plugin.xml
```diff
@@ -297,15 +297,21 @@
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.InsertDefaultPublicConstructorToMBeanQuickFix" />
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidInjectAnnotatedMethodParamAnnotation"
-                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectParamAnnotationQuickFix" />    
+                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectProducesParamAnnotationQuickFix" />
```
mezarin, line 300: Similar to the parameters, instead of renaming the class to capture the cases it handles, perhaps there should be different quick fixes for both of the cases that are being handled: initializer methods and producer methods:
RemoveInvalidIntializerMethodParamAnnotationQuickFix
RemoveInvalidProducerMethodParamAnnotationQuickFix
The code duplication should be minimal since the parent class is the one that does all of the work.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/plugin.xml
```diff
@@ -297,15 +297,21 @@
...
                   targetDiagnostic="jakarta-cdi#InvalidInjectAnnotatedMethodParamAnnotation"
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInjectAnnotationQuickFix" />   
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidProducerMethodParamAnnotation"
-                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectParamAnnotationQuickFix" />     
+                  class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveInvalidInjectProducesParamAnnotationQuickFix" />     
       <codeAction kind="quickfix"
                   targetDiagnostic="jakarta-cdi#InvalidProducerMethodParamAnnotation"
                   class="org.eclipse.lsp4jakarta.jdt.internal.cdi.RemoveProduceAnnotationQuickFix" />
+      <codeAction kind="quickfix"
```
mezarin, line 310: For the added code action that is associated with InvalidDisposesAnnotationOnMultipleMethodParams, what happens if the method has 3+ `@Dispose` annotated parameters? Is only one  removed? Also, it looks like the code action would read incorrectly if the method had only `@Dispose` annotated parameters. Please double check that.

mezarin (commented):
Had some comments.

mezarin (approved):

PR #535, opened 2024-11-07
Remove diagnostic for managed bean public field with invalid scope

### LINKED ISSUE #534: Invalid diagnostic on a field in a managed bean
#### Description:
The fix for issue #444 introduced a new diagnostic but this behaviour is not required by the spec.
image

#### Specification:
The [specification](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#managed_beans) states:
```
If a managed bean has a non-static public field, it [the managed bean] must have scope @Dependent. If a managed bean with a non-static public field declares any scope other than @Dependent, the container automatically detects the problem and treats it as a definition error.
```
We should not generate a diagnostic for a field with a scope declaration. The second sentence in the quote above is not a separate rule, it specifies the container behaviour when the rule of the first sentence is broken.

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

turkeylurkey:
We clarified this rule with the Liberty CDI team and decided the latest added diagnostic is not correct. The fix for the original diagnostic is correct. The issue #534 has a bit more explanation.

https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#managed_beans

Fixes #534


mezarin (approved):
LGTM

PR #536, opened 2024-11-13
Fixes #416 - Quick-fix fails due to getData() is null

### LINKED ISSUE #416: [VS Code] Quick-fix fails due to getData() is null
**To reproduce**
Using [jakarta-sample](https://github.com/eclipse/lsp4jakarta/tree/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample) project in VS Code (Mac OS), trying to trigger quick-fixes on bean validation constraint diagnostics: 

```
[Error - 12:57:13 PM] Request textDocument/codeAction failed.
  Message: Request jakarta/java/codeaction failed with message: java.lang.NullPointerException: Cannot invoke "Object.toString()" because the return value of "org.eclipse.lsp4j.Diagnostic.getData()" is null
  Code: -32603 
```

image

aparnamichael:
Fixes #416 

Quick-fix fails due to getData() is null.
 The root cause of the issue is lsp4Jakarta expects a Map object from VS Code, but we are sending a string or array of strings from VS code. Resolved the null pointer exception by adding code to handle both the 'data' format.

eg:
data = “AssertTrue”
data =[“ApplicationScoped", "RequestScoped”]

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2019, 2023 Red Hat Inc. and others.
+* Copyright (c) 2019, 2024 Red Hat Inc. and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
@@ -23,12 +23,9 @@ import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextDocumentIdentifier;
-import org.eclipse.lsp4jakarta.jdt.internal.beanvalidation.RemoveDynamicConstraintAnnotationQuickFix;
 
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
 
 /**
  * Arguments utilities.
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
@@ -114,17 +114,20 @@ public class ArgumentUtils {
         }
         List<Map<String, Object>> diagnosticsObj = (List<Map<String, Object>>) contextObj.get(DIAGNOSTICS_PROPERTY);
         List<Diagnostic> diagnostics = diagnosticsObj.stream().map(diagnosticObj -> {
+            LOGGER.info("Received diagnostic data" + diagnosticObj.toString());
             Diagnostic diagnostic = new Diagnostic();
             diagnostic.setRange(getRange(diagnosticObj, RANGE_PROPERTY));
             diagnostic.setCode(getString(diagnosticObj, CODE_PROPERTY));
             diagnostic.setMessage(getString(diagnosticObj, MESSAGE_PROPERTY));
             diagnostic.setSource(getString(diagnosticObj, SOURCE_PROPERTY));
             // In Eclipse IDE (LSP client), the data is JsonObject, and in JDT-LS (ex :
-            // vscode as LSP client) the data is a Map, we
-            // convert the Map to a JsonObject to be consistent with any LSP clients.
-            LOGGER.info("diagnosticObj from vscode ============ " + diagnosticObj.toString());
-
-            diagnostic.setData(getObjectAsJson(diagnosticObj, DATA_PROPERTY));
+            // vscode as LSP client) the data is a Map.
+            
+            // In Vscode we are sending data in two different formats either as a string or
+            // as an array of strings. eg: data = “AssertTrue” or  data =[“ApplicationScoped", "RequestScoped”]
+            // if it is a string -> set data as an Object.
+            // if ite is an array of strings - > set data as an JsonArray
+            diagnostic.setData(getValueFromDataParameter(diagnosticObj, DATA_PROPERTY));
             return diagnostic;
         }).collect(Collectors.toList());
         List<String> only = null;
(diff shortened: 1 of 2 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
```diff
@@ -15,16 +15,20 @@
...
 import java.util.Map;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import org.eclipse.lsp4j.CodeActionContext;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextDocumentIdentifier;
+import org.eclipse.lsp4jakarta.jdt.internal.beanvalidation.RemoveDynamicConstraintAnnotationQuickFix;
```
turkeylurkey, line 26: Please remove this unused import.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
```diff
@@ -139,20 +150,32 @@ public static Map<String, Object> getObject(Map<String, Object> obj, String key)
...
+     * @return Returns the child as a JSON array if the data parameter contains an array of strings; otherwise, 
+     * returns it as an object if present, or null if not.
      */
-    public static JsonObject getObjectAsJson(Map<String, Object> obj, String key) {
-        Object child = obj.get(key);
-        if (child != null && child instanceof Map<?, ?>) {
+    public static Object getValueFromDataParameter(Map<String, Object> data, String key) {
+
+        Object child = data.get(key);
+        if (child != null && child instanceof String) {
```
turkeylurkey, line 165: You do not need this null check because child will not be an instance of String if it is null. Please remove the null check.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
```diff
@@ -139,20 +150,32 @@ public static Map<String, Object> getObject(Map<String, Object> obj, String key)
...
+        	// if the value in the 'data' is a string, we string the object.
```
turkeylurkey, line 166: This code does not line up because there is a tab character on the line. Please remove the tab and replace with 4 spaces. 
The same goes for the following line, the two comment lines in the next if block and one comment line in the subsequent else block.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
```diff
@@ -139,20 +150,32 @@ public static Map<String, Object> getObject(Map<String, Object> obj, String key)
...
+        } else if (child instanceof List<?>) {
+        	// if the value in the 'data' is an array, we will convert it to an JsonArray.
+        	// eg: data =[“ApplicationScoped", "RequestScoped”]
             Gson gson = new Gson();
-            return (JsonObject) gson.toJsonTree(child);
+            JsonArray jsonArray = gson.toJsonTree(child).getAsJsonArray();
+            return jsonArray;
+        } else {
+        	//Returns the object if it exists and is an object, and null otherwise
+            return getObject(data, key);
```
turkeylurkey, line 177: Could you clarify why we need this line? It checks if the data is a Map?

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
aparnamichael: done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
aparnamichael: done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
aparnamichael: done

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
aparnamichael: I was expecting the 'getObject' method to return an object, as its return type is 'Map<String, Object>. Could I replace it with 'return null'?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
```diff
@@ -139,20 +147,34 @@ public static Map<String, Object> getObject(Map<String, Object> obj, String key)
...
+	 * @param key  the key of the child
+	 * @return Returns the child as a JSON array if the data parameter contains an
+	 *         array of strings; otherwise,
+	 *         returns it as an object if present, or null if not.
+	 */
+    public static Object getValueFromDataParameter(Map<String, Object> data, String key) {
+
+    	Object child = data.get(key);
+    	if (child instanceof String) {
+    		// if the value in the 'data' is a string, we string the object.
```
turkeylurkey, line 165: Unfortunately it appears you added a tab to every line rather than change to spaces. Could you configure your editor to use spaces instead of tabs? Alternatively could you use a different editor like vi that will make it more obvious when you are using tabs and give you control over whether tabs or spaces are used?

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
turkeylurkey: It would probably return null all the time anyway, thanks.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/ls/ArgumentUtils.java
aparnamichael: @turkeylurkey Please check now, I have corrected the indentation issue.

PR #539, opened 2024-11-14
Remove a quick fix that removes both annotations when the two are mutually exclusive

turkeylurkey:
Pull request #511 added a quick fix which removes all three annotations from a parameter when some of the annotations are in conflict. E.g. "@Disposes @Observes @ObservesAsync String name" There should be two quick fixes, one that removes @Disposes and one that removes @Observes and @ObservesAsync. 

In this pull request we temporarily remove the quick fix from two situations along with the tests that validate them. There is an issue open to fix the code so we can restore the quick fix.


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/plugin.xml
ajm01, line 1: LGTM

PR #553, opened 2025-03-25
Prepare 0.2.5

turkeylurkey:
Update release number for next release. We are skipping `0.2.4` due to an error in the Eclipse repository: we accidentally pushed an invalid file using that version number and the Maven system has no "un-do" function.


mezarin (approved):
The changes look good. I understand that there was an issue that triggered 0.2.4 to be partially released in error, and that is the reason why we are skipping it here.
Approving if there is no way to delete the invalid 0.2.4 release/data from maven central.

turkeylurkey:
I worked with the Eclipse help desk but they also agree there is no way to fix this.

PR #558, opened 2025-07-14
Multiple @BeanParams incorrectly reported as error

Rejoice472:
Issue link = https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/522

As per https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#entity_parameters
BeanParam is not an entity parameter as it used to map other resources like @queryParam and @pathParam etc.
Therefore, it can be excluded from the ‘only one entity parameter’ validation rule.

I can raise a PR for IntelliJ once this change is confirmed.


turkeylurkey:
I agree with your fix. Could you add a unit test? `org.eclipse.lsp4jakarta.jdt.test`

turkeylurkey (approved):
Thank you for the fix.

PR #562, opened 2025-07-24
Fix to address JavaCursorContextKind IllegalArgumentException

archana-1924:
The PR addresses Issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520
 Root causes were observed in JavaCursorContextKind.java files in 
>lsp4jakarta.ls
>lsp4jakarta.jdt

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
@@ -226,6 +226,8 @@ public enum JavaCursorContextKind {
 
     public static JavaCursorContextKind forValue(int value) {
         JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
+        //Root cause for https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520. See issue for more details.
+        //Fix for Issue 520 lsp4jakarta author: archana-1924
         if (value == NONE.getValue()) {
             return allValues[allValues.length - 1];
         } else {
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2023 Red Hat Inc. and others.
+* Copyright (c) 2023, 2025 Red Hat Inc. and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
@@ -224,17 +224,15 @@ public enum JavaCursorContextKind {
         return value;
     }
 
-    public static JavaCursorContextKind forValue(int value) {
-        JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
-		
-        // Root cause and fix for https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520.
-		// See issue for more details.		
+	public static JavaCursorContextKind forValue(int value) {
+		JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
+
+		// Root cause and fix for JavaCursorContextKind IllegalArgumentException issue #520 when value passed is 2000
 		if (value == NONE.getValue()) {
 			return NONE;
-		} else {
-			if (value < 1 || value > allValues.length)
-				throw new IllegalArgumentException("Illegal enum value: " + value);
-			return allValues[value - 1];
 		}
-    }
+		if (value < 1 || value > allValues.length-1)
+			throw new IllegalArgumentException("Illegal enum value: " + value);
+		return allValues[value - 1];
+	}
 }
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
@@ -224,15 +224,15 @@ public enum JavaCursorContextKind {
         return value;
     }
 
-	public static JavaCursorContextKind forValue(int value) {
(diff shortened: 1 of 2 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
```diff
@@ -226,8 +226,14 @@ public int getValue() {
...
         JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
-        if (value < 1 || value > allValues.length)
-            throw new IllegalArgumentException("Illegal enum value: " + value);
-        return allValues[value - 1];
+        //Root cause for https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520. See issue for more details.
+        //Fix for Issue 520 lsp4jakarta author: archana-1924
+        if (value == NONE.getValue()) {
+            return allValues[allValues.length - 1];
+        } else {
+	        if (value < 1 || value > allValues.length)
```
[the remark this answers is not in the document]
turkeylurkey: It is best to change the settings of your editor not to use tabs for paragraphing.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
```diff
@@ -226,8 +226,15 @@ public int getValue() {
 
     public static JavaCursorContextKind forValue(int value) {
         JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
-        if (value < 1 || value > allValues.length)
-            throw new IllegalArgumentException("Illegal enum value: " + value);
-        return allValues[value - 1];
+		
+        // Root cause and fix for https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520.
```
turkeylurkey, line 230: Please update the comment to describe the reason for the if stmt rather than refer to the issue number. This makes it easier for someone else to understand the code because they do not need to go look up the issue.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
```diff
@@ -226,8 +226,15 @@ public int getValue() {
...
+		// See issue for more details.		
+		if (value == NONE.getValue()) {
+			return NONE;
+		} else {
+			if (value < 1 || value > allValues.length)
```
turkeylurkey, line 235: Since we have already handled the item at the end of the array this check is off by one. The check should be against `(allValues.length - 1)`. As it stands now the value 9 would be allowed and return NONE even though it only goes from 1-8 + 2000 (NONE).

turkeylurkey (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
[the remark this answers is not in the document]
turkeylurkey: I don't see the tabs converted to spaces yet.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
```diff
@@ -226,9 +226,16 @@ public int getValue() {
 
     public static JavaCursorContextKind forValue(int value) {
         JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
-        if (value < 1 || value > allValues.length)
-            throw new IllegalArgumentException("Illegal enum value: " + value);
-        return allValues[value - 1];
+        
+        //Root cause and fix for https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/520.
+        //See issue for more details.
```
Rejoice472, line 231: I think you can avoid the nested if with something like this.

`if (value == NONE.getValue()) {
    return NONE;
}
if (value < 1 || value > allValues.length) {
    throw new IllegalArgumentException("Illegal enum value: " + value +
        ". Expected between 1 and " + allValues.length);
}
return allValues[value - 1];`

Rejoice472 (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
archana-1924: Yeah, that's a great catch. Thanks, will think about other corner scenarios too.

jakarta.ls/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
turkeylurkey: The latest code removes the `else` which is great.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/commons/JavaCursorContextKind.java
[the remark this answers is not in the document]
turkeylurkey: All the other issues are resolved and the code looks great except for this issue. I think it is important to maintain the standard to use 4 spaces to indent each level. Normally it is simple to replace tabs with 4 spaces but I understand that your usual editor may automatically change it back. Could you please find an editor that will allow you to make these changes and let you check it in? Sometimes I use the editor `vi` on MacOS for things like this.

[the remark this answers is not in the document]
archana-1924: Changed the indentation using vi editor. Please check if it's okay

turkeylurkey (approved):
Looks good.

archana-1924:
> @archana-1924 If your pull request comment says "Fixes: XXX" where XXX is the issue number then when we merge the pull request github will automatically close the issue. Also, remember you may wish to offer the same pull request in LSP4MP following largely the same process. Just follow their rule about the name of the pull request.

Sure thing. lsp4mp was in my radar too, will do the same.

Issue #507, opened 2024-01-04
Investigate whether LSP4Jakarta consistently processes fully qualified class names.


mrglavas:
- Check whether LSP4Jakarta has the correct behaviour for applications if they specify a fully qualified class name instead of their simple name (e.g. `@jakarta.ws.rs.PUT` vs `@PUT`).
- Check whether LSP4Jakarta unambiguously matches the expected class names when checking diagnostics, applying quick fixes, etc... (e.g. `@jakarta.ws.rs.PUT` vs `@my.random.pkg.PUT`).

Note: We're noticing inconsistencies in the processing of qualified names in the code actions / quick fixes in Liberty Tools for IntelliJ and that code was ported from LSP4Jakarta.

mrglavas:
A couple more tasks for this item:

- [ ] Check for usage of [DiagnosticUtils.getSimpleName()](https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/DiagnosticUtils.java). Investigate how it's being used to shorten qualified names and determine if that usage is problematic or may become a problem in the future as Jakarta EE evolves.
- [ ] Review the diagnostic collectors, checking if there are any other methods being used for shortening qualified names.

mrglavas:
Should check whether the issues (in Liberty Tools for IntelliJ) with qualified names being tracked here: https://github.com/OpenLiberty/liberty-tools-intellij/issues/879 also apply to this project.

Rejoice472:
Issues observed so far and corresponding prs.
1.)  https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/567
https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/569
2.) https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/571
https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/572
3.) https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/574
https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/576

cc: @turkeylurkey , @TrevCraw

turkeylurkey:
Ok, thank you.

PR #565, opened 2025-07-29
Adding back the quick fixes to plugin.xml

archana-1924:
This PR addresses issue from LSP4Jakarta: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/540

PR has fixes in: 
JDT element: Adds back the quick fixes which were temporarily commented because it produced quick fixes which removed mutually exclusive annotations.

Added 2 new quick fixes which gives separate fixes for mutually exclusive annotations.

Diagnostics targeted:
image
image


Rejoice472 (approved):

turkeylurkey:
Two things: I think part of this work is to remove `RemoveInvalidInitializerParamAnnotationQuickFix` because it removes all three annotations whereas the only problem is that some of them are mutually exclusive. So we're really splitting that quick fix into two. Are you able to test this code in the Eclipse IDE? That way you can see how the quick fixes look in the list the UI presents it is easier to see anomalies like the overlapping function of the QF listed above and the new QFs you are adding.

This code uses simple names. Issue #507 states that we need to be using fully qualified names. So the quick fixes need to work the same whether the code uses simple names or fully qualified names. This problem is outside the scope of this issue but could you work with @Rejoice472 and see if issue #567 will cover the FQ names aspect of this? If not could you open an issue for the FQ problem of these quick fixes? Also somehow organize the issues related to quick fixes using simple names. E.g. an Epic that collects them into a list.

turkeylurkey (changes requested):
Added a comment to the pull request.

archana-1924:
Fully qualified names DEMO:
image
image

archana-1924:
@turkeylurkey : All the comments are addressed and tested. FQ names and normally addressed are working fine for the diagnostic.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/RemoveInvalidDisposerConflictParamAnnotationQuickFix.java
```diff
@@ -0,0 +1,46 @@
...
+ * Removes the @Observes and @ObservesAsync annotations from
+ * the declaring element.
+ */
+public class RemoveInvalidDisposerConflictParamAnnotationQuickFix extends RemoveMethodParamAnnotationQuickFix {
+
+    /**
+     * Constructor.
+     */
+    public RemoveInvalidDisposerConflictParamAnnotationQuickFix() {
+        super(Constants.INVALID_DISPOSER_FQ_CONFLICTED_PARAMS.toArray((String[]::new)));
```
Rejoice472, line 28: Why do we need double conversion efforts? It would be better to declare INVALID_DISPOSER_FQ_CONFLICTED_PARAMS as an array directly, which would avoid redundant conversions—especially if that constant isn't used elsewhere.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/RemoveInvalidDisposerParamAnnotationQuickFix.java
```diff
@@ -16,31 +16,31 @@
...
-public class RemoveInvalidInitializerParamAnnotationQuickFix extends RemoveMethodParamAnnotationQuickFix {
+public class RemoveInvalidDisposerParamAnnotationQuickFix extends RemoveMethodParamAnnotationQuickFix {
 
     /**
      * Constructor.
      */
-    public RemoveInvalidInitializerParamAnnotationQuickFix() {
-        super(Constants.INVALID_INITIALIZER_PARAMS_FQ);
+    public RemoveInvalidDisposerParamAnnotationQuickFix() {
+        super(Constants.INVALID_DISPOSER_FQ_PARAMS.toArray((String[]::new)));
```
Rejoice472, line 28: Why do we need double conversion efforts? It would be better to declare INVALID_DISPOSER_FQ_PARAMS as an array directly, which would avoid redundant conversions—especially if that constant isn't used elsewhere.

archana-1924: Done

Rejoice472:
> @turkeylurkey : All the comments are addressed and tested. FQ names and normally addressed are working fine for the diagnostic.
> 
> Had an observation during unit testing, for the @Inject diagnostic, the fully qualified param names are not appearing on the diagnostic, while for the **disposer** diagnostic, the fully qualified param names are shown in the message. We may have to keep the behavior standard whether to show just the name or fully qualified name. However, this won't be part of the scope of this issue. Can discuss if it could be done as part of FQ name issue #567 @Rejoice472

Hi @archana-1924 , Thanks for finding this. I hope the label will always be the short name. cc: @turkeylurkey 
I've addressed this in the existing open PR.
https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/579/files

turkeylurkey (approved):
Looks good.

turkeylurkey:
The only thing is that some other quick fixes seem to appear twice. I suspect they are created by other diagnostics and are not related to this. Could you investigate and see if they appear properly and consider if there is something you can do? For instance, if the quick fixes are due to different diagnostics could you modify the message so that different quick fixes seem different? On other platforms quick fixes are shown with a string that indicates the source but not on Eclipse.
image

PR #569, opened 2025-08-01
Code Action to Remove Annotation Fails with Fully Qualified Annotation Names 

Rejoice472:
Code Action to Remove Annotation Fails with Fully Qualified Annotation Names. Ideally, it should be able to remove the specific annotation even when the fully qualified name is used.
The same functionality works in IntelliJ.
defect link -> https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/567
This defect was identified as part of the investigation I performed while analysing https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/507

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/RemoveDynamicConstraintAnnotationQuickFix.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/RemoveDynamicConstraintAnnotationQuickFix.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2023 IBM Corporation and others.
+* Copyright (c) 2023, 2025 IBM Corporation and others.
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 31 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/RemoveDynamicConstraintAnnotationQuickFix.java
```diff
@@ -115,7 +115,9 @@ public CodeAction resolveCodeAction(JavaCodeActionResolveContext context) {
      *
      * @return The code action label.
      */
-    private static String getLabel(String annotationName) {
+    private static String getLabel(String annotation) {
+        String[] parts = annotation.split("\\.");
```
arunvenmany-ibm, line 119: can we simplify the logic like this?

String annotationName = annotation.contains(".") ? annotation.substring(annotation.lastIndexOf('.') + 1) : annotation;

arunvenmany-ibm (changes requested):
Verify whether the copyright years are updated in the source files.

Usually the copyright will have a start_year,last_modified_year
if it has only start_year add a modified year as well

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/jsonb/JsonbDiagnosticsParticipant.java
```diff
@@ -171,8 +171,7 @@ private boolean createJsonbTransientDiagnostic(JavaDiagnosticsContext context, S
         } else if (errorCode.equals(ErrorCode.InvalidJSonBindindAnnotationWithJsonbTransientOnAccessor)) {
             diagnosticErrorMessage = Messages.getMessage("ErrorMessageJsonbTransientOnAccessor");
         }
-        // convert to simple name for current tests
-        List<String> diagnosticData = jsonbAnnotations.stream().map(annotation -> DiagnosticUtils.getSimpleName(annotation)).collect(Collectors.toList());
+        List<String> diagnosticData = jsonbAnnotations.stream().collect(Collectors.toList());
```
turkeylurkey, line 174: I don't understand this line. It converts a List<String> to a List<String>. Formerly it converted the names. Is this line still here because we need to make a copy of the list to pass to Gson? Or can it be removed? N.B. this construction appears a few times.

turkeylurkey: 👍

turkeylurkey (approved):

PR #575, opened 2025-08-06
New test case for Type check and Method naming conventions : Issue-532

Joseph-Bineesh:
New test cases added for issue https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/532
1. Test the return type of method/ type of field
2. Test the naming convention of getter method and access specifier


Joseph-Bineesh:
Mockito is not accepted as solution. So closing this PR.

Issue #577, opened 2025-08-06
System is not able to detect the fully qualified name and giving wrong diagnosis


Rejoice472:
#### Description:
System is not able to detect the fully qualified name and giving wrong diagnosis
<img width="877" height="275" alt="Image" src="https://github.com/user-attachments/assets/b49e713e-4b4d-4577-a158-8bd1880d5aed" />

#### Type of language feature proposed:
_Select all that apply_
- [ ] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

archana-1924:
https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/565 PR for adding new quick fixes

PR #579, opened 2025-08-08
The system is unable to detect the fully qualified name and is providing incorrect diagnostics.

Rejoice472:
defect link = #577 
This defect was identified as part of the investigation I performed while analysing #507 
Addressing both scenarios which mentioned in the 577


arunvenmany-ibm (approved):

turkeylurkey (approved):
Looks good

Issue #580, opened 2025-08-11
The default value for the type annotation attribute should be Object.class, not an empty string.


Rejoice472:
#### Description:
The default value of the type attribute of @Resource annotation should be Object.class rather than an empty string. This discrepancy causes the quick fix to fail during compilation.

<img width="450" height="63" alt="Image" src="https://github.com/user-attachments/assets/eb2afade-9406-47ba-8d11-09ae76f48112" />

#### Type of language feature proposed:
_Select all that apply_
- [ ] diagnostic
- [x] quick-fix
- [ ] snippet
- [ ] other, please specify:

turkeylurkey:
Normally I find it more useful to read this part of the spec: https://jakarta.ee/specifications/annotations/2.0/annotations-spec-2.0.html#jakarta-annotation-resource.
In this case the extra context was not needed, thanks.

PR #581, opened 2025-08-11
The default value for the type annotation attribute should be Object.class, not an empty string.

Rejoice472:
This is to fix defect #580


Joseph-Bineesh (approved):
LGTM

turkeylurkey (approved):
Looks good but could you check if the IntelliJ quick fix has the same problem and open an issue if required? Thanks.

Issue #192, opened 2021-10-21
Dependency Injection: Diagnostics for non-static inner class managed bean


KidoVin01:
[About CDI Managed Beans](https://eclipse-ee4j.github.io/jakartaee-tutorial/#about-cdi-managed-beans)
> A top-level Java class is a managed bean if it is defined to be a managed bean by any other Jakarta EE technology specification, such as the Jakarta Faces technology specification, or if it meets all the following conditions.
> - It is not a nonstatic inner class.

### Diagnostic
1. Identify the field/method/constructor annotated with `@Inject`
2. Check if the dependency class is contained in the same file as an inner class
3. If the dependency class is NOT `static`, deliver a diagnostic error that the dependency class must be `static`

### Quick Fix
- Remove `@Inject` annotation
- Suggest adding `static` modifier to class declaration

### [Example](https://github.com/AlvinTan2000/demo-jakarta/blob/main/src/main/java/io/openliberty/sample/jakarta/di/unsatisfied/stat/GreetingServlet.java)
The following is a faulty code snippet example
```java
@WebServlet(name = "diUnStat", urlPatterns = { "/diUnStat" })
public class GreetingServlet extends HttpServlet {
    private static final long serialVersionUID = 1L;
        
    @Inject
    private GreetingA greeting;

    @Override
	protected void doGet(HttpServletRequest req, HttpServletResponse res) throws ServletException, IOException {
        res.setContentType("text/html;charset=UTF-8");
		res.getWriter().println(greeting.greet("Bob"));
	}
    
    // Missing `static` modifier
    @ApplicationScoped
    public class GreetingA {
    	public String greet(String name) {
            return "GreetingA, " + name;
        }
    }
}

```

<br />

```
[INFO] [ERROR   ] CWWKZ0004E: An exception occurred while starting the application demo-servlet. The exception message was: com.ibm.ws.container.service.state.StateChangeException: org.jboss.weld.exceptions.DeploymentException: WELD-001408: Unsatisfied dependencies for type GreetingA with qualifiers @Default
[INFO]   at injection point [BackedAnnotatedField] @Inject private io.openliberty.sample.jakarta.di.unsatisfied.stat.GreetingServlet.greeting
[INFO]   at io.openliberty.sample.jakarta.di.unsatisfied.stat.GreetingServlet.greeting(GreetingServlet.java:0)
```

<br />

Related to #153

PR #596, opened 2025-09-09
Dependency Injection: Diagnostics for non-static inner class managed bean

Rejoice472:
Added test cases to support the fix in PR https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/595 for issue #192


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/test/di/InnerClassInjectionTest.java
```diff
@@ -0,0 +1,77 @@
+/*******************************************************************************
+ * Copyright (c) 2021, 2025 IBM Corporation and others.
```
arunvenmany-ibm, line 2: when a new class is created, only start year( ie, 2025) is required

arunvenmany-ibm (approved):
Looks good

Issue #284, opened 2022-04-05
[New feature title]


giancarlopernudisegura:
#### Description:
In #258, if you have a server endpoint path that does not start with a leading slash, you get a diagnostic message. As pointed out by @kathrynkodama, it would be beneficial to have a quick-fix that adds the leading slash.

#### Specification:
> All server endpoint paths must start with a leading `/`

[source](https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0.html#uri-mapping)

#### Type of language feature proposed:
_Select all that apply_
- [ ] diagnostic
- [x] quick-fix
- [ ] snippet
- [ ] other, please specify:

PR #604, opened 2025-09-17
Issue 284: Add Quick-Fix for ServerEndpoint paths without a leading slash

Joseph-Bineesh:
#284 Quick fix to prefix with '/' for ServerEndpoint  annotation value.


arunvenmany-ibm (approved):
looks good

Issue #633, opened 2025-11-06
Diagnostic for missing public no-arg constructor for @ServerEndpoint and @ClientEndpoint classes


Joseph-Bineesh:
#### Description:
Classes annotated with @ServerEndpoint or @ClientEndpoint must define a public no-argument constructor. The runtime uses this constructor to instantiate endpoint instances. If a public no-arg constructor is missing, the endpoint cannot be created, causing deployment or runtime failures. 

#### Diagnostic Purpose: 
This diagnostic identifies endpoint classes that cannot be instantiated due to the absence of a public no-argument constructor. It ensures compliance with the WebSocket specification and proper runtime behavior. 

#### Typical Causes: 

- Defining only constructors with parameters. 
- Defining a non-public no-argument constructor (e.g., private or protected). 
- Omitting the constructor entirely when another constructor is defined (the default public no-arg constructor is no longer provided). 

#### Specification:
https://jakarta.ee/specifications/websocket/2.0/websocket-spec-2.0#serverendpoint

#### Type of language feature proposed:
_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

PR #795, opened 2026-01-12
PR: Diagnostic for missing public no-arg constructor for @ServerEndpoint and @ClientEndpoint classes

Joseph-Bineesh:
#633

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/websocket/DefaultConstructorTest.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/websocket/DefaultConstructorTest.java
@@ -1,0 +1,9 @@
+package io.openliberty.sample.jakarta.websocket;
+
+import jakarta.websocket.server.ServerEndpoint;
+
+@ServerEndpoint("/path")
+public class DefaultConstructorTest {
+
+	
+}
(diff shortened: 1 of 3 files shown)


Joseph-Bineesh:
https://github.com/user-attachments/assets/0daa5a14-f921-4419-801a-836917c1fcd8

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/websocket/InsertDefaultPublicConstructorQuickFix.java
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
+* Copyright (c) 2023 IBM Corporation and others.
```
venmanyarun: change copyright year

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/websocket/WebSocketDiagnosticsParticipant.java
```diff
@@ -363,6 +366,33 @@ private void serverEndpointErrorCheck(JavaDiagnosticsContext context, String uri
...
+                hasConstructor = true;
+                String[] params = method.getParameterTypes();
+                int flags = method.getFlags();
+                if (params.length == 0 && Flags.isPublic(flags)) {
+                    hasPublicNoArgConstructor = true;
+                }
+            }
+        }
+
+        if (hasConstructor && !hasPublicNoArgConstructor) {
```
venmanyarun: why we need to check hasConstructor here? could there be a scenario where hasConstructor=false and hasPublicNoArgConstructor is true?

venmanyarun (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/websocket/WebSocketDiagnosticsParticipant.java
Joseph-Bineesh: hasConstructor becomes false only in the case of class only have default constructor(public no arg).  But it will not counted in type.getMethods(). In this case hasPublicNoArgConstructor also has value false. Irrespective of values we should not show any diagnostics if default constructor exist.

venmanyarun (approved):

Rejoice472 (approved):

PR #810, opened 2026-02-13
Fixes #489: Automating data sync up between jakarta-sample and demo-servlet-no-diagnostics

### LINKED ISSUE #489: Enhancement to dynamically use test project for "no diagnostics" automated tests
#### Description:
In https://github.com/eclipse/lsp4jakarta/pull/483, a new test project was added which includes fake Jakarta imports in order to test the scenario where we should only be providing diagnostics when the fully qualified Jakarta package name is available as an import.

Currently, this means that any updates made to the original test project will need to be made to the second project as well. In order to avoid dual maintenance, we could dynamically retrieve the Java files from the original test project ([jakarta-sample](https://github.com/eclipse/lsp4jakarta/tree/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample)) and then modify the imports to make them fake before passing them as input to the "no-diagnostics" test.

archana-1924:
Fixes issue #489 

Here, we can pass the arguments for SOURCE and DESTINATION directories, to point to specific set of test cases and generate copies of those test cases in demo-servlet-no-diagnostics. If not passed, it will default run it on all the test cases present in jakarta-sample.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -67,7 +67,7 @@ The following instructions explain how to set up your Eclipse IDE workspace.
 
         <img src="/docs/images/building_lsp4e_4.png" alt="lsp4e project build path confirming jar" height="60%" width="60%"/>
 
-## Commit process
+## Autogenerate No diagnostics test cases
 
 1. Add your code changes and dataset addition/modification used for testing in jakarta-sample project.
 
@@ -78,16 +78,6 @@ The following instructions explain how to set up your Eclipse IDE workspace.
     2. Don't pass any arguments: bash syncTestDatasets.sh - Takes the latest changes for all the java files inside jakarta-sample project and syncs it with demo-servlet-no-diagnostics project with fake imports.
 
 3. bash buildAll.sh - Checks if project builds fine.
-
-4. Stage the changes and put an appropriate commit message.
-
-5. Do a signed commit for the verified commit to appear in github.
-
-6. Merge and raise PR.
-
-7. Once the reviewers have resolved comment(s) if any and approved PR, it will be merged by the committers.
-
-8. Close the associated issue in the repository.
 
 ## Common Errors 
 


syncDemoJakartaSample.sh
```diff
@@ -0,0 +1,49 @@
...
+
+  # Normalize destination file by stripping fake, for comparison
+  if [ -f "$destFile" ]; then
+    sed -E 's/import[[:space:]]+jakarta\.fake\./import jakarta./g' "$destFile" > "$tmpDest"
+  fi
+
+  # Compare normalized files
+  if [ ! -f "$destFile" ] || ! cmp -s "$tmpSrc" "$tmpDest"; then
+    echo "Updating $destFile"
+    cp "$srcFile" "$destFile"
```
Joseph-Bineesh: Is it replace entire file?

Joseph-Bineesh: ok

syncDemoJakartaSample.sh
Joseph-Bineesh: Add a README for how to execute this script file and if you can find a better file name would be good.

Joseph-Bineesh (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/demo-servlet-no-diagnostics/src/main/java/io/openliberty/sample/jakarta/jaxrs/MultipleEntityParamsResourceMethod.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2021 IBM Corporation.
```
venmanyarun: correct copyright year

PR #812, opened 2026-02-17
Feature/default attr values

### LINKED ISSUE #611: Default values for annotation attributes should be determined by their actual type.
This PR: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/581 attempted to fix https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/580 by simply checking whether an annotation attribute name is `type` and assumes that it can assign `Object.class` as a default value. The code which existed prior to this PR assumed that all annotation attributes can be assigned an empty string as a default value.

The current design is not taking into account the actual type of the annotation attribute and will produce code that will not compile because the default value chosen is not actually assignable. A declaration for the annotation being processed should be available and should be introspected to determine the type of the annotation attribute.

The allowed types for annotation attributes are:
- Primitive types: `byte`, `short`, `int`, `long`, `float`, `double`, `boolean`, `char`.
- String: The `java.lang.String` class.
- Class: The `java.lang.Class` type, or any invocation of Class (e.g., Class<T>).
- Enum types: Any `enum` type declared in Java.
- Annotation types: Other annotation types. This allows for nested annotations.
- Arrays of any of the preceding types: Single-dimension arrays of any of the above-mentioned types (e.g., `int[]`, `String[]`, `Class<?>[]`).

Once the type has been determined, a legal assignable value should be used as a default value. For example for a primitive type `0` or `false` (for a boolean) could be a default value. If the type is actually a String, use `""`. If the type is actually a Class, use `Object.class`. For an enum, perhaps just arbitrarily choose one of the constants and assign that as the default value.

Rejoice472:
fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/611

https://github.com/user-attachments/assets/a61073a9-ae77-477e-8ea5-d9271768a4e8

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/ModifyAnnotationProposal.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/ModifyAnnotationProposal.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021, 2025 IBM Corporation and others.
+ * Copyright (c) 2021, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -29,8 +29,6 @@ import org.eclipse.jdt.core.dom.CompilationUnit;
 import org.eclipse.jdt.core.dom.Expression;
 import org.eclipse.jdt.core.dom.FieldDeclaration;
 import org.eclipse.jdt.core.dom.IBinding;
-import org.eclipse.jdt.core.dom.IMethodBinding;
-import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.eclipse.jdt.core.dom.MemberValuePair;
 import org.eclipse.jdt.core.dom.NormalAnnotation;
 import org.eclipse.jdt.core.dom.SingleMemberAnnotation;
@@ -323,7 +321,7 @@ public class ModifyAnnotationProposal extends InsertAnnotationProposal {
             if (annotationToProcess == null) {
                 // We are adding a new required default @Resource annotation to an empty
                 // @Resources annotation.
-                addNewAttributes(ast, values, annotationToProcess);
+                addNewAttributes(ast, values, annotation, annotationToProcess);
             } else {
                 // get the existing name/value pairs from the existing NormalAnnotation that was
                 // passed into this method above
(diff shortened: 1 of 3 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/ModifyAnnotationProposalHelper.java
```diff
@@ -0,0 +1,231 @@
...
+     */
+    private static Expression createCustomDefaultValue(AST ast, ITypeBinding typeBinding) {
+        if (typeBinding == null) {
+            return ast.newNullLiteral();
+        }
+        if (typeBinding.isArray()) {
+            return ast.newArrayInitializer();
+        }
+        if (typeBinding.isEnum()) {
+            for (IVariableBinding field : typeBinding.getDeclaredFields()) {
```
venmanyarun: I dont think this is correct. getDeclaredFields() does not guarantee order, this code will return the first enum constant it finds, which might not be the first one written in the source code.
may be sort based on start position and then choose the first enum constant

venmanyarun (changes requested):
update tests as well for 
Class<?>
java.lang.Class
Class<? extends Foo>

Multi-dimensional arrays (String[][])
Generic arrays (List[])

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/ModifyAnnotationProposal.java
```diff
@@ -393,32 +389,26 @@ private NormalAnnotation createNewAnnotation(AST ast, ImportRewrite imports, Imp
...
 
     }
 
-    private List<MemberValuePair> addNewAttributes(AST ast, List<MemberValuePair> values) {
-        // Add new attributes of type String or Class.
-        // For initial values, we use empty strings for String types and Object.class for Class types,
-        // since the user's intended values are unknown at this stage,
-        // These placeholders (e.g., name = "", type = Object.class) must be updated by the user as needed.
-        // when an annotation in Jakarta EE declares an attribute named type, it’s always of the form of Class<?>
+    private List<MemberValuePair> addNewAttributes(AST ast, List<MemberValuePair> values, String annotationFqn, NormalAnnotation annotationToProcess) {
```
Joseph-Bineesh: Add documentation for method

Joseph-Bineesh (commented):
Can we add dedicated test cases for these changes?

Rejoice472:
> We currently support three quick‑fix scenarios: Object.class, string values, and arrays. Tests for these are updated. Other cases are future‑oriented; if such scenarios arise, the logic will handle them, and we can add quick fixes with tests then.

We currently support three quick‑fix scenarios: Object.class, string values, and arrays. Tests for these are updated. Other cases are future‑oriented; if such scenarios arise, the logic will handle them, and we can add quick fixes with tests then.

venmanyarun (approved):

venmanyarun (approved):

Joseph-Bineesh (approved):

PR #924, opened 2026-07-14
Fixes #693 Diagnostics for component class which declares or inherits class-level interceptor binding must be declared appropriately.

### LINKED ISSUE #693: Component class which declares or inherits class-level interceptor binding must be declared appropriately.
#### Description:
If a component class declares or inherits a class-level interceptor binding, it must not be declared final, or have any non-static, non-private, final methods. If a component has a class-level interceptor binding and is declared final or has a non-static, non-private, final method, the container automatically detects the problem and treats it as a definition error and causes deployment to fail. 

#### Diagnostic Purpose:  
Interceptor bindings enable cross-cutting concerns (e.g., logging, security, transactions) to be applied to component classes. For the container to apply interceptors, it often uses dynamic proxying or subclassing. The presence of final modifiers interferes with this: 
* A final class cannot be subclassed. 
* A component class carries a class-level interceptor binding but declares a method that is `final` and neither `static` nor `private`. 

This leads to: 
* Deployment failure 
* Silent omission of interceptors 
* Broken application behavior 

#### Typical Causes:  
Component class which declares or inherits class-level interceptor binding are declared final, or have any non-static, non-private, final methods. 

#### Specification:  
A component class carries a class-level interceptor binding but declares a method that is `final` and neither `static` nor `private`. CDI proxies must be able to override business methods to weave in interceptor logic; a `final` method blocks that. The container must treat this as a definition error.
https://jakarta.ee/specifications/interceptors/2.0/interceptors-spec-2.0#:~:text=If%20a%20component%20class%20declares%20or%20inherits%20a%20class%2Dlevel%20interceptor%20binding%2C%20it%20must%20not%20be%20declared%20final%2C%20or%20have%20any%20non%2Dstatic%2C%20non%2Dprivate%2C%20final%20methods 

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify:

archana-1924:
Fixes #693 

#### Valid Code

```java
@Logged
@ApplicationScoped
public class ReportService {

    // ✅ Regular public method — fine
    public String generate(String type) { return "report:" + type; }

    // ✅ private final — exempt (private methods are excluded by the spec)
    private final String sanitize(String input) { return input.trim(); }

    // ✅ static final — exempt (static methods are excluded by the spec)
    public static final String version() { return "1.0"; }
}
```

#### Invalid Code

```java
@Logged
@ApplicationScoped
public class ReportService {

    // ❌ public final — "non-static, non-private, final" → DefinitionException
    public final String generate(String type) { return "report:" + type; }

    // ❌ protected final — also "non-static, non-private, final" → DefinitionException
    protected final String formatTitle(String title) { return "[" + title + "]"; }
}
```


archana-1924:
https://github.com/user-attachments/assets/9375f504-a0d0-4818-95ee-af99b3bf1e87

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/interceptor/InvalidInterceptorModifiersOnClassMethod.java
```diff
@@ -0,0 +1,17 @@
...
+import jakarta.interceptor.Interceptors;
+
+@Interceptors(Monitored.class)
+public final class InvalidInterceptorModifiersOnClassMethod {
+	// Diagnostic: non-static, non-private final method
+	public final void processPayment() {
+		
+    }
+	// No diagnostic: private final method is permitted
+	private final void internalProcess() {
```
Rejoice472, line 18: could you please add all combinations of static, private, public and final since the rule says 
 it must not be declared final, or have any non-static, non-private, final methods.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/interceptor/InterceptorDiagnosticsParticipant.java
```diff
@@ -421,4 +423,71 @@ private void checkInterceptorBinding(IType type, ICompilationUnit unit, String u
...
+        }
+        if (Flags.isFinal(typeFlag)) {
+            Range range = PositionUtils.toNameRange(type, context.getUtils());
+            diagnostics.add(context.createDiagnostic(uri,
+                                                     Messages.getMessage("InvalidFinalInterceptorBindingClass"),
+                                                     range, Constants.DIAGNOSTIC_SOURCE,
+                                                     ErrorCode.InvalidFinalInterceptorBindingClass,
+                                                     DiagnosticSeverity.Error));
+        }
+        for (IMethod method : type.getMethods()) {
```
Joseph-Bineesh, line 447: checkClassLevelInterceptorBindingConstraints method intend for class level check. Can we move method  level check to any existing method loop ?

Joseph-Bineesh (changes requested):

sajeerzeji (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/interceptor/InterceptorDiagnosticsParticipant.java
archana-1924: This is kept separate for the binding logic for the reason that we don't check if it's @Interceptor annotated which is the top level check for the current parent loop. This doesn't need it and works on 2 other annotations on type level.

Joseph-Bineesh: I see. That means still it checks the rules for class level in methods inside class?

Joseph-Bineesh (approved):
LGTM

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/interceptor/InterceptorDiagnosticsParticipant.java
archana-1924: It was named checkClassLevelInterceptorBindingConstraints because Interceptor Bindings are given at class level. Renamed for readability.

Rejoice472 (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/interceptor/InterceptorBindingClassModifiers.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // Valid: private static final method is permitted
+    private static final void privateStaticHelper() {
+    }
+}
+
+// VALID: no final modifier on class or non-private/non-static methods - no diagnostic expected
+@Monitored
+@Interceptor
+class ValidInterceptorBindingClassModifiers {
```
venmanyarun: this should produce zero diagnostics — please add a test asserting that

venmanyarun (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/interceptor/InterceptorBindingClassModifiers.java
archana-1924: It was inside InterceptorBindingClassModifier which has invalid scenarios. So, not having any diagnostics was the indication that it was a positive scenario. To validate it separately, added this to a separate class and wrote the Junit.

venmanyarun (approved):

venmanyarun (approved):

PR #928, opened 2026-07-15
Fix annotation removal collapsing line break and overlapping text edits

### LINKED ISSUE #927: Quick fix "Remove @[Scope]" collapses line break between annotation and class declaration
# Description

When applying a quick fix to remove a conflicting scope annotation from a class that has two scope annotations on the same line, the fix incorrectly removes the newline between the annotation line and the class declaration, merging them onto a single line.

This was originally reported in [OpenLiberty/liberty-tools-vscode#390](https://github.com/OpenLiberty/liberty-tools-vscode/issues/390).

# Steps to reproduce

1. Create a Java file with the following code:

``` java
 @ApplicationScoped @RequestScoped
public class ScopeDeclaration {

} 
```

2. Apply the quick fix "Remove @RequestScoped"

Expected result:

``` java
@ApplicationScoped
public class ScopeDeclaration {

}
```

Actual result:

``` java
@ApplicationScoped public class ScopeDeclaration {

}
```

# Root cause

The bug is in RemoveAnnotationProposal.getRewrite(). The fix uses ASTRewrite.remove(annotationNode, null) to remove the annotation node from the modifiers list. When JDT's rewriter removes a node that is the last token on a line, it extends the delete range forward past the \n character to clean up surrounding whitespace, which collapses the annotation line and the class declaration line into one.

# Why the fix must live in this repo

The TextEdit (the delete range) is computed entirely inside RemoveAnnotationProposal in org.eclipse.lsp4jakarta.jdt.core. By the time it reaches the language client (Liberty Tools for VS Code or any other client), it is just a range + empty string with no context about line boundaries. The client has no way to know the range is wrong so  the fix must be applied at the source where the range is computed.

Proposed fix:

Override addEdits() in RemoveAnnotationProposal instead of using getRewrite(). Compute a precise DeleteEdit directly from the raw source text with two cases:

1. Annotation is alone on its line. In this case, delete the entire line including its trailing \n (no blank line left behind)
2. Annotation shares its line with other tokens → delete the annotation plus preceding horizontal whitespace only, with a hard stop at the line boundary so the \n is never consumed

A working implementation with this fix has been validated across all cases (same line at end, same line at start, alone on own line) in Liberty Tools for VS Code.

sharma1208:
Fixes #927 

# Problem
When applying "Remove @Annotation" on a class with two scope annotations on the same line, the line break between the annotation and public class was incorrectly eaten. When three annotations shared a line and two were removed simultaneously, the operation failed with Overlapping text edits.

# Fix
Replaced getRewrite() with an addEdits() override that computes precise DeleteEdit ranges from raw source text, guaranteed to never cross a line boundary or overlap with adjacent ranges.

# Test Cases Working: 

https://github.com/user-attachments/assets/ae9c9b1f-6cab-4aaf-9a1a-1fccb858a3b0

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
@@ -124,7 +124,7 @@ public class RemoveAnnotationProposal extends ASTRewriteCorrectionProposal {
                 continue;
             }
             ITypeBinding binding = annotation.resolveTypeBinding();
-            if (!binding.getQualifiedName().equals(matchingFqn)) {
+            if (binding == null || !binding.getQualifiedName().equals(matchingFqn)) {
                 continue;
             }
 
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
@@ -195,14 +195,19 @@ public class RemoveAnnotationProposal extends ASTRewriteCorrectionProposal {
                         }
                         scan--;
                     }
-                    // If the preceding line is itself an annotation, do not absorb backward —
-                    // absorbing into another annotation's line would corrupt that annotation.
+                    // If the preceding line is an annotation or a comment, do not absorb
+                    // backward — absorbing into an annotation would corrupt it, and
+                    // absorbing into a comment would produce confusing output.
                     if (prevLineHasContent) {
                         int prevLineStart = prevLineEnd - 1;
                         while (prevLineStart > 0 && source.charAt(prevLineStart - 1) != '\n') {
                             prevLineStart--;
                         }
-                        if (source.substring(prevLineStart, prevLineEnd).stripLeading().startsWith("@")) {
+                        String prevLineContent = source.substring(prevLineStart, prevLineEnd).stripLeading();
+                        if (prevLineContent.startsWith("@")
+                                || prevLineContent.startsWith("//")
+                                || prevLineContent.startsWith("/*")
+                                || prevLineContent.startsWith("*")) {
                             prevLineHasContent = false;
                         }
                     }
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
@@ -13,6 +13,7 @@
 *******************************************************************************/
 package org.eclipse.lsp4jakarta.jdt.core.java.corrections.proposal;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -109,8 +110,10 @@ public class RemoveAnnotationProposal extends ASTRewriteCorrectionProposal {
         }
 
         String source = document.get();
+        List<int[]> pendingDeletes = new ArrayList<>();
 
-        for (ASTNode child : children) {
+        for (int i = 0; i < children.size(); i++) {
+            ASTNode child = children.get(i);
             if (!(child instanceof Annotation)) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


venmanyarun:
@sharma1208 

Many JDT.LS tests are failing.. please check

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
```diff
@@ -89,45 +93,191 @@ protected ASTRewrite getRewrite() throws CoreException {
...
-                    String matchingFqn = Arrays.stream(annotations).filter(fqn -> matchesAnnotation(fqn, annotation.getTypeName().toString())).findFirst().orElse(null);
-                    if (matchingFqn != null) {
-                        // Resolving fully qualified name from Annotation to fix issue #567
-                        ITypeBinding binding = annotation.resolveTypeBinding();
-                        if (binding.getQualifiedName().equals(matchingFqn)) {
-                            rewrite.remove(child, null);
+
+            // Find the end of the line containing this annotation (exclusive of \n).
+            int lineEnd = annotEnd;
+            while (lineEnd < source.length() && source.charAt(lineEnd) != '\n') {
```
venmanyarun: do not hardcode \n. use lineseparator instead

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
```diff
@@ -73,7 +73,11 @@ public RemoveAnnotationProposal(String label, ICompilationUnit targetCU, Compila
     }
 
     @Override
-    protected ASTRewrite getRewrite() throws CoreException {
+    protected void addEdits(IDocument document, TextEdit editRoot) throws CoreException {
+        // Call CUCorrectionProposal.addEdits() (empty default), not
+        // ASTRewriteCorrectionProposal.addEdits(), since we handle edits directly.
+        super.addEdits(document, editRoot);
```
venmanyarun: The PR overrides getRewrite() to return null. But the parent class ASTRewriteCorrectionProposal.addEdits() still calls getRewrite(). The PR also overrides addEdits() and calls super.addEdits(document, editRoot) which is ASTRewriteCorrectionProposal.addEdits(). That method calls getRewrite(), receives null, skips the rewrite — which is fine. But the original non-overridden fallback path in ASTRewriteCorrectionProposal.getRewrite() throws:

```
IStatus status = StatusFactory.newErrorStatus("Rewrite not initialized", null);
throw new CoreException(status);
```
This path is only hit if fRewrite == null, which it will be here since the constructor passes null as the rewrite argument. The new getRewrite() override returns null, so ASTRewriteCorrectionProposal.addEdits() receives null and safely skips the edit — this is currently safe only because the if (rewrite != null) guard exists in the parent. 

However, the comment in the PR says "Not used — addEdits handles everything directly to avoid ASTRewrite producing delete ranges that cross line boundaries." This is misleading: the purpose of overriding getRewrite() here is just to suppress the error thrown by the grandparent when fRewrite == null. 

The correct fix is to either not call super.addEdits() (skip the grandparent chain entirely) or skip the getRewrite() override entirely and simply not call super.addEdits(). As written, the dead override adds confusion and a latent risk if the parent changes.

Recommendation: Remove the getRewrite() override. Instead, in addEdits(), do NOT call super.addEdits(), and separately handle import rewrites directly. This is cleaner and avoids the entire class hierarchy ambiguity.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
```diff
@@ -89,45 +93,191 @@ protected ASTRewrite getRewrite() throws CoreException {
...
+                    }
+                    deleteEnd = nextLineStart + nextLineIndent;
                 }
             }
 
-            return rewrite;
+            pendingDeletes.add(new int[]{deleteStart, deleteEnd});
         }
 
+        // Merge overlapping or adjacent DeleteEdits so that removing multiple
```
venmanyarun, line 256: Add a log warning when pendingDeletes is empty at the end (annotations expected to be removed but none found). This is not a crash bug but a silent failure.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
```diff
@@ -89,45 +93,191 @@ protected ASTRewrite getRewrite() throws CoreException {
...
+            } else {
+                editRoot.addChild(new DeleteEdit(current[0], current[1] - current[0]));
+                current = new int[]{range[0], range[1]};
+            }
+        }
+        if (current != null) {
+            editRoot.addChild(new DeleteEdit(current[0], current[1] - current[0]));
+        }
+
+        // Rewrite imports if the import rewrite has accumulated changes.
```
venmanyarun, line 275: Add tests for import rewrites into existing test cases

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
```diff
@@ -89,45 +93,191 @@ protected ASTRewrite getRewrite() throws CoreException {
...
+            }
+        }
+    }
+
+    /**
+     * Not used — {@link #addEdits} handles everything directly to avoid
+     * ASTRewrite producing delete ranges that cross line boundaries.
+     */
+    @Override
+    protected ASTRewrite getRewrite() throws CoreException {
```
venmanyarun: As mentioned above, remove overriding getRewrite

venmanyarun (changes requested):
For tests, I can see that you have done trial and error in each commit to fix tests..

You should run buildAll.sh in local before committing changes, to see all tests are successful

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
sharma1208: Thanks for the detailed explanation, you are definitely correct. The getRewrite() override was only there to suppress the crash from the uninitialized fRewrite, and super.addEdits() was calling back into the parent's import flush while we were also doing it manually at the bottom. Both have been removed. Imports are now handled directly in addEdits() only, and getRewrite() is gone.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/core/java/corrections/proposal/RemoveAnnotationProposal.java
sharma1208: Used TextUtilities.getDefaultLineDelimiter(document) instead of System.lineSeparator() since this reads the delimiter from the document itself rather than the host OS, which seems correct when files may have different line endings than the system default.

PR #948, opened 2026-07-31
Fixes #684 Diagnostic logic for Invalid passivating scopes

### LINKED ISSUE #684: A managed bean in a passivating scope must implement Serializable
#### Description:
In CDI, a passivating scope (such as @SessionScoped or @ConversationScoped) allows beans to be passivated and restored across requests. To support this, the bean must implement Serializable. If a managed bean declares a passivating scope but does not implement Serializable, the container must treat it as a definition error. 

This diagnostic flags beans in passivating scopes that do not implement Serializable 

#### Specification:
https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#passivation_validation

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify: 

#### Invalid example: 

```
@SessionScoped 
public class UserSession { 
    private String name; 
} 
//  Invalid: @SessionScoped requires Serializable 
```

Valid example: 

```
@SessionScoped 
public class UserSession implements Serializable { 
    private String name; 
} 
// Valid: bean is serializable 
```

archana-1924:
Fixes #684


archana-1924:
https://github.com/user-attachments/assets/4c2d13e5-4ecf-4cb0-aedd-3085061e6660

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/test/cdi/ManagedBeanTest.java
```diff
@@ -110,7 +114,17 @@ public void ManagedBeanWithDependent() throws Exception {
                           "The @Dependent annotation must be the only scope defined by a Managed bean class of generic type.",
                           DiagnosticSeverity.Error, "jakarta-cdi", "InvalidGenericManagedBeanClassWithNoDependentScope");
 
-        assertJavaDiagnostics(diagnosticsParams, IJDT_UTILS, d1, d2, d3, d4, d5, d6);
+        Diagnostic d7 = d(17, 6, 27,
```
Joseph-Bineesh: Add a proper diagnostics variable names

Joseph-Bineesh (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/src/main/java/org/eclipse/lsp4jakarta/jdt/test/cdi/ManagedBeanTest.java
archana-1924: This is pending as it requires discussion with team, since entire Unit tests are in a changed state now.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
```diff
@@ -506,6 +517,94 @@ public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProg
...
+     * @param uri the URI of the compilation unit
+     * @param diagnostics the list to add diagnostic errors to
+     * @param type the Java type being validated
+     * @param unit the compilation unit
+     * @throws JavaModelException if there is an error accessing Java model elements
+     */
+    private void validatePassivatingScopeWithoutSerializable(JavaDiagnosticsContext context, String uri,
+                                                             List<Diagnostic> diagnostics, IType type,
+                                                             ICompilationUnit unit) throws JavaModelException {
+        if (!hasPassivatingScope(type, unit)) {
```
Joseph-Bineesh: Do we really need to check again hasPassivatingScope? This is already checked in calling method.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/core/java/ManagedBean.java
```diff
@@ -373,4 +373,45 @@ public static boolean hasMetaAnnotation(IAnnotation annotation, IType type, ICom
...
+     * @param metaAnnotationFQN the fully qualified name of the meta-annotation to find
+     * @param memberName the name of the member whose value to return
+     * @param memberType the expected type class of the member value
+     * @return the member value, or {@code null} if not found
+     * @throws JavaModelException if there is an error accessing Java model elements
+     */
+    public static <T> T getMetaAnnotationMemberValue(IAnnotation annotation, IType type, ICompilationUnit cu,
+                                                     String metaAnnotationFQN, String memberName,
+                                                     Class<T> memberType) throws JavaModelException {
+        if (!hasMetaAnnotation(annotation, type, cu, metaAnnotationFQN)) {
```
Rejoice472, line 399: Why are we calling hasMetaAnnotation here? The code inside hasMetaAnnotation is used afterwards, and getChildITypeByName (a heavy operation) ends up running twice. Can we avoid this?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
```diff
@@ -425,6 +426,15 @@ public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProg
                 }
             }
 
+            // A managed bean (or custom-scoped class) in a passivating scope must implement Serializable.
+            // CDI 3.0 §6.6.4: Passivating scopes must be declared @NormalScope(passivating=true).
+            // Built-in passivating scopes: @SessionScoped and @ConversationScoped.
+            // Custom passivating scopes: any annotation meta-annotated with @NormalScope(passivating=true).
+            // https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#passivating_scopes
+            if (hasPassivatingScope(type, unit)) {
```
Rejoice472, line 434: As per the spec, what I understand is that this check should be inside one of the isManagedBean checks, correct?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
```diff
@@ -506,6 +516,89 @@ public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProg
...
+     *
+     * @param annotation the annotation present on the bean class
+     * @param type the Java type being validated (used for name resolution)
+     * @param unit the compilation unit (used for import-aware matching of binary types)
+     * @return {@code true} if the annotation represents a custom passivating scope
+     * @throws JavaModelException if there is an error accessing Java model elements
+     */
+    private boolean isCustomPassivatingScope(IAnnotation annotation, IType type,
+                                             ICompilationUnit unit) throws JavaModelException {
+        return Boolean.TRUE.equals(ManagedBean.getMetaAnnotationMemberValue(
```
Rejoice472, line 597: This is a suggestion.

The current hasPassivatingScope chain (isCustomPassivatingScope → getMetaAnnotationMemberValue → hasMetaAnnotation → getChildITypeByName → resolveType + findType) repeats type resolution unnecessarily; can we mirror the existing validateInterceptorDecoratorScopes() pattern, which uses getFullyQualifiedClassName + findType with caching to solve the same problem in one pass.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
archana-1924: Current ManagedBean check only checks for built in scopes and not the custom ones. For our case, we need the custom scopes along with conversation and session scoped, which we have explicitly checked in our case.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
archana-1924: After the pre-flight fix, the chain is now: isCustomPassivatingScope → getMetaAnnotationMemberValue → getChildITypeByName → getFullyQualifiedClassName (= resolveType) + findType. So, both paths do the same resolveType + findType. 

The actual remaining difference - we cannot simply swap to the validateinterceptorDecoratorScopes pattern as the validateinterceptorDecoratorScopes pattern deliberately only needs presence (any @NormalScope is invalid on an interceptor/decorator). The passivating scope check must read the passivating attribute — a scope annotated with @NormalScope(passivating=false) must not trigger the Serializable diagnostic.

Joseph-Bineesh (approved):
LGTM. 

Suggestion to refactor method collectDiagnostics onto sub methods for better readability.

Rejoice472 (approved):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/ManagedBeanDiagnosticsParticipant.java
```diff
@@ -488,6 +498,89 @@ public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProg
...
+     * <p>Checks both built-in passivating scopes ({@code @SessionScoped},
+     * {@code @ConversationScoped}) and custom scopes whose annotation type is
+     * meta-annotated with {@code @NormalScope(passivating=true)}.
+     *
+     * @param type the Java type being validated
+     * @param unit the compilation unit
+     * @return {@code true} if any passivating scope annotation is present
+     * @throws JavaModelException if there is an error accessing Java model elements
+     */
+    private boolean hasPassivatingScope(IType type, ICompilationUnit unit) throws JavaModelException {
```
Rejoice472: Can we use a stream here so that it stays consistent with the LTI implementation?

archana-1924: Because the isCustomPassivatingScope throws JavaModelException, if we wrap this also in streams using try/catch, compared to the current for-loop, it becomes:

More verbose — two extra try/catch blocks just to bridge checked→unchecked→checked
Changes nothing semantically — the exception still propagates the same way
Is harder to read — the intent is buried in boilerplate

PR #952, opened 2026-08-04
Legal bean types

### LINKED ISSUE #667: Flag use of type variables as bean types
#### Description:
This diagnostic identifies cases where a type variable (e.g., T, E, K) is incorrectly used as a bean type in CDI.
According to the CDI specification, type variables are not legal bean types because they do not represent concrete, instantiable types at runtime.

#### Specification:
https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#legal_bean_types
https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0#producer_method 

#### Rules:

1. Type variables are not legal bean types. Any injection point, bean class, or producer method using a type variable (T, E, etc.) is invalid.
2. Producer methods must declare concrete types. If the return type is a parameterised type, all type parameters must be specified with actual types.
If the return type is a type variable, or an array whose component type is a type variable, the container treats it as a definition error.
3. Parameterised types with missing type arguments are invalid. Example: List without <String> or another concrete type argument is not allowed.

_Select all that apply_
- [x] diagnostic
- [ ] quick-fix
- [ ] snippet
- [ ] other, please specify: 

#### Invalid Examples:
```
@ApplicationScoped
public class MyBean<T> {

    @Inject
    T value;  

    // T is a type variable — not a legal bean type
}
```
```
@Produces
List produceList();  

// Missing type argument — List without a concrete type parameter is not a legal bean type

```
```
@Produces
<T> T produceGeneric();  

// Return type is a type variable — definition error

```
```
@Produces
T[] produceArray();  

// Array component type is a type variable — definition error
```

#### valid Examples:
```
@ApplicationScoped
public class MyBean {

    @Inject
    String value;  

    // Concrete type — legal bean type
}
```
```
@Produces
List<String> produceList() {
    return new ArrayList<>();
}

// Parameterized type with actual type argument — valid

```
```
@Produces
Map<String, Integer> produceMap() {
    return new HashMap<>();
}

// All type parameters specified with concrete types — valid

```

Rejoice472:
fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/667

https://github.com/user-attachments/assets/699c3ce3-e8db-462c-a85b-51b9e8fb827a


sajeerzeji (approved):
LGTM

PR #968, opened 2026-08-31
Diagnostic typeuse constraints

Rejoice472:
fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/619

https://github.com/user-attachments/assets/96cf24ef-9de0-4d55-96c7-6569af0f722c

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
@@ -63,6 +63,11 @@ import org.eclipse.jdt.core.IMethod;
 import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.Signature;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+import org.eclipse.jdt.core.dom.IAnnotationBinding;
+import org.eclipse.jdt.core.dom.ITypeBinding;
+import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
+import org.eclipse.lsp4jakarta.jdt.core.ASTUtils;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
 import org.eclipse.lsp4j.Range;
@@ -104,6 +109,9 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
         IAnnotation[] annotations;
         IMethod[] allMethods;
 
+        // Build AST once for TYPE_USE traversal (type argument annotations)
+        List<FieldDeclaration> fieldDeclarations = ASTUtils.getFieldDeclarations(unit);
+
         alltypes = unit.getAllTypes();
         for (IType type : alltypes) {
             allFields = type.getFields();
@@ -120,6 +128,8 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
                         validAnnotation(context, uri, field, annotation, matchedAnnotation, diagnostics);
                     }
                 }
+                // Check TYPE_USE annotations on generic type arguments of this field
+                checkTypeUseAnnotations(context, uri, field, type, fieldDeclarations, diagnostics);
             }
             allMethods = type.getMethods();
             for (IMethod method : allMethods) {
@@ -187,7 +197,7 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
                 case ASSERT_FALSE, ASSERT_TRUE -> {
                     String message = getDiagnosticMessage(isMethod, isField, annotationName, "AnnotationBoolean");
 
-                    if (!type.equals(Signature.SIG_BOOLEAN) && !getDataTypeName(type).equals("Boolean")) {
+                    if (!type.equals(Signature.SIG_BOOLEAN) && !DiagnosticUtils.getDataTypeName(type).equals("Boolean")) {
                         Range range = PositionUtils.toNameRange(element, context.getUtils());
                         diagnostics.add(context.createDiagnostic(uri, message, range, Constants.DIAGNOSTIC_SOURCE,
                                                                  matchedAnnotation, ErrorCode.InvalidAnnotationOnNonBooleanMethodOrField,
@@ -196,7 +206,7 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
                 }
                 case DECIMAL_MAX, DECIMAL_MIN, DIGITS -> {
                     String dataTypeFQName = DiagnosticUtils.getMatchedJavaElementName(declaringType,
-                                                                                      getDataTypeName(type),
+                                                                                      DiagnosticUtils.getDataTypeName(type),
                                                                                       NUMERIC_AND_CHAR_WRAPPER_TYPES);
 
                     if (dataTypeFQName == null && !type.equals(Signature.SIG_BYTE)
@@ -213,7 +223,7 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
                 }
                 case EMAIL, NOT_BLANK, PATTERN -> checkStringOnly(context, uri, element, diagnostics, annotationName, isMethod, type, matchedAnnotation, declaringType, isField);
                 case FUTURE, FUTURE_OR_PRESENT, PAST, PAST_OR_PRESENT -> {
-                    String dataType = getDataTypeName(type);
+                    String dataType = DiagnosticUtils.getDataTypeName(type);
                     String dataTypeFQName = DiagnosticUtils.getMatchedJavaElementName(declaringType, dataType,
                                                                                       SET_OF_DATE_TYPES.toArray(new String[0]));
                     if (dataTypeFQName == null) {
@@ -226,7 +236,7 @@ public class BeanValidationDiagnosticsParticipant implements IJavaDiagnosticsPar
                 }
                 case MIN, MAX -> {
                     String dataTypeFQName = DiagnosticUtils.getMatchedJavaElementName(declaringType,
-                                                                                      getDataTypeName(type),
+                                                                                      DiagnosticUtils.getDataTypeName(type),
                                                                                       NUMERIC_WRAPPER_TYPES);
                     if (dataTypeFQName == null && !type.equals(Signature.SIG_BYTE)
                         && !type.equals(Signature.SIG_SHORT) && !type.equals(Signature.SIG_INT)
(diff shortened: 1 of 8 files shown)


jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
```diff
@@ -104,6 +109,9 @@ public List<Diagnostic> collectDiagnostics(JavaDiagnosticsContext context, IProg
         IAnnotation[] annotations;
         IMethod[] allMethods;
 
+        // Build AST once for TYPE_USE traversal (type argument annotations)
+        List<FieldDeclaration> fieldDeclarations = ASTUtils.getFieldDeclarations(unit);
```
Joseph-Bineesh, line 113: Is it for getting the type of fields?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
```diff
@@ -291,6 +303,158 @@ private void validAnnotation(JavaDiagnosticsContext context, String uri, IJavaEl
...
+     * @param type the declaring IType (used for name resolution)
+     * @param fieldDeclarations pre-collected field declarations from the AST root
+     * @param diagnostics the list to add diagnostics to
+     */
+    private void checkTypeUseAnnotations(JavaDiagnosticsContext context, String uri, IField field,
+                                         IType type, List<FieldDeclaration> fieldDeclarations,
+                                         List<Diagnostic> diagnostics) throws CoreException {
+        String fieldName = field.getElementName();
+
+        for (FieldDeclaration node : fieldDeclarations) {
```
Joseph-Bineesh, line 325: Can we directly resolve field type and check the it is valid? Any particular case why we use ASTVisit?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: Not Only the type type with generics

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: The JDT model API (IField.getTypeSignature()) returns the erased type signature — e.g. for List<@NotNull String> it returns "QList;". Type parameters are stripped. There is no way to retrieve the type arguments or their annotations through the JDT model API.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: Not only the type, but also decorated types such as generics.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: We need AST traversal to detect annotations inside type generics.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/resources/org/eclipse/lsp4jakarta/jdt/core/messages/messages.properties
```diff
@@ -97,6 +97,14 @@ AnnotationMinMaxParams = The {0} annotation can only be used on \n- BigDecimal \
...
+AnnotationBooleanTypeUse = The {0} annotation can only be used on boolean and Boolean type arguments.
+AnnotationBigDecimalTypeUse = The {0} annotation can only be used on: \n- BigDecimal \n- BigInteger \n- CharSequence\n- byte, short, int, long (and their respective wrappers) \n type arguments.
+AnnotationStringTypeUse = The {0} annotation can only be used on String and CharSequence type arguments.
+AnnotationDateTypeUse = The {0} annotation can only be used on: Date, Calendar, Instant, LocalDate, LocalDateTime, LocalTime, MonthDay, OffsetDateTime, OffsetTime, Year, YearMonth, ZonedDateTime, HijrahDate, JapaneseDate, MinguoDate and ThaiBuddhistDate type arguments.
+AnnotationMinMaxTypeUse = The {0} annotation can only be used on \n- BigDecimal \n- BigInteger\n- byte, short, int, long (and their respective wrappers) \n type arguments.
+AnnotationPositiveTypeUse = The {0} annotation can only be used on \n- BigDecimal \n- BigInteger\n- byte, short, int, long, float, double (and their respective wrappers) \n type arguments.
+SizeOrNonEmptyAnnotationsTypeUse = This annotation can only be used on type arguments of type CharSequence, Collection, Array, or Map.
```
Joseph-Bineesh: Can we follow the same pattern 'The {0} annotation can only' for this message as well?

Joseph-Bineesh (changes requested):

Joseph-Bineesh (approved):
LGTM

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
```diff
@@ -291,6 +303,158 @@ private void validAnnotation(JavaDiagnosticsContext context, String uri, IJavaEl
...
+                try {
+                    String matched = DiagnosticUtils.getMatchedJavaElementName(declaringType, annFQN,
+                                                                               SET_OF_ANNOTATIONS.toArray(new String[0]));
+                    if (matched == null || matched.equals(VALID))
+                        continue;
+
+                    // erasedName: strip type parameters (e.g. "java.util.List<X>" → "java.util.List")
+                    String qualifiedName = resolved.getQualifiedName();
+                    int paramStart = qualifiedName.indexOf('<');
+                    String erasedName = paramStart >= 0 ? qualifiedName.substring(0, paramStart) : qualifiedName;
```
archana-1924: Can you see if we can use resolved.getErasure().getQualifiedName() instead of manual slicing?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
```diff
@@ -291,6 +303,158 @@ private void validAnnotation(JavaDiagnosticsContext context, String uri, IJavaEl
...
+                                                      annBinding.getAnnotationType().getName(),
+                                                      "SizeOrNonEmptyAnnotationsTypeUse",
+                                                      ErrorCode.InvalidAnnotationOnNonSizeTypeUse);
+                            }
+                        }
+                        default -> {
+                            /* not a constraint annotation — skip */ }
+                    }
+                } catch (CoreException e) {
+                    LOGGER.log(Level.WARNING, "Error checking TYPE_USE annotation on field " + field.getElementName(), e);
```
archana-1924, line 431: Is the scope limited to only fields?

How about below use cases?
Method return types: List<@NotNull String> 
Method parameters: void getNames(List<@NotNull String> names)
Record components / Constructor parameters

archana-1924 (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
```diff
@@ -291,6 +303,158 @@ private void validAnnotation(JavaDiagnosticsContext context, String uri, IJavaEl
...
+
+                    // erasedName: strip type parameters (e.g. "java.util.List<X>" → "java.util.List")
+                    String qualifiedName = resolved.getQualifiedName();
+                    int paramStart = qualifiedName.indexOf('<');
+                    String erasedName = paramStart >= 0 ? qualifiedName.substring(0, paramStart) : qualifiedName;
+
+                    // Each case checks validity and emits the TYPE_USE-specific error code.
+                    // Primitives are not valid generic type arguments (Java compile error), so
+                    // only wrapper/object types need to be checked here.
+                    switch (matched) {
```
archana-1924, line 365: Do you want to consider extracting common logic for the below as we have an existing switch case similar to this?
The switch (matched) block duplicates the same annotation compatibility checks as field/method validation in BeanValidationDiagnosticsParticipant.
To improve maintainability when new constraints or supported types are added, consider extracting a helper method (e.g. checkConstraintTypeCompatibility(String matchedAnnotation, String typeFqn, IType declaringType)) shared between regular element validation and type argument validation.

archana-1924 (changes requested):

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: Yes, this was initially discussed even before development. However, due to complexity and performance concerns, the scope has been limited to fields only.

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/beanvalidation/BeanValidationDiagnosticsParticipant.java
Rejoice472: I looked into extracting a shared helper as you suggested. The two switch blocks look similar on the surface but operate on fundamentally different type representations:

validAnnotation works on raw JDT Signature strings ("Z", "I", "[B", …) and must short-circuit on Signature.SIG_BOOLEAN, SIG_BYTE, etc. to allow primitives where they're valid (e.g. @Min on an int field).
validateTypeArgBindings works on erased FQNs from ITypeBinding.getErasure().getQualifiedName() ("java.lang.Boolean", "java.math.BigInteger", …). Primitives can never appear as type arguments so there are no primitive short-circuits.
Sharing the compatibility logic required introducing a third data structure (TYPE_USE_DIAGNOSTIC_INFO map) to replace the inner switch that maps annotations to (messageKey, ErrorCode) pairs — which actually adds more complexity than it removes. The original two focused switch blocks are more straightforward to read and maintain.

PR #972, opened 2026-09-04
Feature/delegate non injection check

Rejoice472:
fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/918

https://github.com/user-attachments/assets/342a9d1e-ae0c-4132-924f-88d187076603

### DIFF
--- a/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/CdiDecoratorDiagnosticsParticipant.java
+++ b/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/CdiDecoratorDiagnosticsParticipant.java
@@ -100,7 +100,7 @@ public class CdiDecoratorDiagnosticsParticipant implements IJavaDiagnosticsParti
 
         List<IJavaElement> delegateElements = new ArrayList<>();
         for (IField field : type.getFields()) {
-            validateDelegate(type, field, field, uri, context, diagnostics, delegateElements);
+            validateDelegate(type, field, field, uri, context, diagnostics, delegateElements, false);
         }
         for (IMethod method : type.getMethods()) {
             IAnnotation[] methodAnnotations = method.getAnnotations();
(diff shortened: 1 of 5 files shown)


Joseph-Bineesh (approved):
LGTM

archana-1924:
In the video, we saw the @InsertInject is coming up with another existing diagnostic? Is that expected?

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/CdiDecoratorDiagnosticsParticipant.java
```diff
@@ -121,19 +138,25 @@ private void validateDecorator(IType type, ICompilationUnit unit, String uri,
...
+                                  IAnnotation... methodAnnotations) throws JavaModelException {
 
         IAnnotation[] annotations = (element instanceof IAnnotatable) ? ((IAnnotatable) element).getAnnotations() : new IAnnotation[0];
 
-        if (DiagnosticUtils.isMatchedAnnotation(type.getCompilationUnit(), annotations, Constants.DELEGATE_FQ_NAME)) {
-            delegateElements.add(element);
-            validateDelegateInjectionPoint(owner,
-                                           methodAnnotations.length > 0 ? methodAnnotations : annotations,
-                                           type, uri, context, diagnostics);
+        if (!DiagnosticUtils.isMatchedAnnotation(type.getCompilationUnit(), annotations, Constants.DELEGATE_FQ_NAME)) {
```
archana-1924, line 153: When a non-void or static method parameter has @Delegate and lacks @Inject:
validateDelegateInjectionPoint hits !isMatchedAnnotation(...) first and reports InvalidDelegateInjectionPoint (which proposes the "Insert @Inject" quickfix).
If the user applies the quickfix "Insert @Inject", the method becomes an @Inject non-void/static method, which is still invalid and triggers diagnostic. Could you please check this scenario?

archana-1924 (changes requested):

Rejoice472:
> In the video, we saw the @InsertInject is coming up with another existing diagnostic? Is that expected?

Yes that is expected

jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/cdi/CdiDecoratorDiagnosticsParticipant.java
Rejoice472: Yes, that's true. We have two options here: show both diagnostics at once or show them one by one. Since Eclipse displays only one diagnostic at a time anyway, I chose the one-by-one approach.

